PR #18, opened 2020-12-14
Fixed certik audit issues

brightdev33:
- Locked compiler version to 0.5.17
- Fixed inefficient storage layout
- Updated comparison inefficient storage read
- Fixed inefficient local variable
- Updated function visibility
- Fixed documentation discrepancy

### DIFF
--- a/contracts/BEP20Interface.sol
+++ b/contracts/BEP20Interface.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.16;
+pragma solidity 0.5.17;
 
 interface BEP20Interface {
   /**
--- a/contracts/BEP20Interface.sol
+++ b/contracts/BEP20Interface.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.5.17;
+pragma solidity ^0.5.16;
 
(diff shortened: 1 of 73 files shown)


contracts/BEP20Interface.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.16;
+pragma solidity 0.5.17;
```
superdev810, line 1: you consider this update will be compatible with current deployed contract?

superdev810 (changes requested):

contracts/BEP20Interface.sol
brightdev33: current contracts are compiled and deployed with 0.5.17, that's why no problem for these contracts.
but need to confirm proxy delegate calls and storage usages for these update.

gryincent, line 1: if so, it needs almost contracts to be redeployed

gryincent (changes requested):

gryincent (approved):

PR #28, opened 2021-03-08
Add vai liquidation and fee module

gryincent:


QichangZhang:
Why omg is used in pledge？

gryincent:
> 
> 
> Why omg is used in pledge？

where is omg you are saying?

zljl22:
Why failed five times

gryincent:
> 
> 
> Why failed five times

it's not related to main update and just testcase params configuration

zljl22:
It seems that the test results are good. ALL test cases have passed. Can you increase some memory parameters to prevent memory overflow?

PR #42, opened 2021-05-06
Implemented Governor Bravo

brightdev33:
- Implemented Governor Bravo
- Added test cases & scenarios


contracts/Governance/GovernorBravoDelegate.sol
```diff
@@ -0,0 +1,388 @@
...
+    uint public constant MIN_VOTING_DELAY = 1;
+
+    /// @notice The max setable voting delay
+    uint public constant MAX_VOTING_DELAY = 40320; // About 1 week
+
+    /// @notice The number of votes in support of a proposal required in order for a quorum to be reached and for a vote to succeed
+    uint public constant quorumVotes = 600000e18; // 600,000 = 2% of Xvs
+
+    /// @notice The maximum number of actions that can be included in a proposal
+    uint public constant proposalMaxOperations = 10; // 10 actions
```
gryincent, line 33: @brightdev33 you can update proposal max operations to 50 in order to handle all markets governance configs in one proposal?

gryincent (changes requested):

contracts/Governance/GovernorBravoDelegate.sol
brightdev33: @gryincent yeah, sure. I tried that already and faced an issue on related tcs and actions. I will update all after fix the issue. thanks for catch.

PR #57, opened 2021-07-12
WIP: Add contribution guidelines

kkirka:
I propose we add some contribution guidelines to the repo so that both the team and the external contributors can see our rules and follow them.

This PR lists the guidelines how I see it but ofc these are up for discussion – feel free to comment or propose the new rules if you think I've missed something.


gryincent:
Thank you. @kkirka 
We can add more if things needed later on for the contributions.

PR #63, opened 2021-08-25
Split supply/borrow XVS rewards

kkirka:
## Description

We need to set separate XVS speeds for borrow and supply (mostly because of the excessive XVS borrowing issue but also because this adds flexibility our risk team needs to adjust the rewards params).

There are also some bugs in the reward distribution mechanism that we inherited from Compound (see the corresponding [bug report](https://www.comp.xyz/t/comptroller-compspeed-bug/2111)).

This PR splits `venusSpeeds` into separate `venusSupplySpeeds` and `venusBorrowSpeeds` and fixes the distribution bugs by moving the market initialization into a separate internal function. The PR is loosely based on https://github.com/compound-finance/compound-protocol/pull/144, though some design decisions are different.

Docs, tests, and simulations are in progress, hence WIP.

## Checklist


kkirka:
Waiting for the vault to make the comptroller smaller

PR #88, opened 2021-12-06
Add Liquidator contract

kkirka:
## Description

Problem: We should split the liquidation incentive between liquidators and treasury. To do this, we need to either upgrade all vTokens (which is not possible due to vBNB), or come up with an intermediary contract that pretends to be a liquidator, and then splits the received vTokens in the right proportion.

Solution: This commit implements the second approach. Liquidator contract is supposed to be the only address allowed to liquidate borrows.

## Checklist

### DIFF
--- a/contracts/ComptrollerInterface.sol
+++ b/contracts/ComptrollerInterface.sol
@@ -85,6 +85,7 @@ interface IVAIVault {
 }
 
 interface IComptroller {
+    function liquidationIncentiveMantissa() external view returns (uint);
     /*** Treasury Data ***/
     function treasuryAddress() external view returns (address);
     function treasuryPercent() external view returns (uint);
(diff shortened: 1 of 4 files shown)


contracts/Comptroller.sol
```diff
@@ -501,6 +504,10 @@ contract Comptroller is ComptrollerV5Storage, ComptrollerInterfaceG2, Comptrolle
         // Shh - currently unused
         liquidator;
```
kkirka, line 505: It's used now, no need to silence the warning

kkirka: Fixed in [a35afda](https://github.com/VenusProtocol/venus-protocol/pull/88/commits/a35afda6e607cb7c155f96ef065128349c1ccf1a)

RoseRompkxm (approved):

PR #90, opened 2021-12-17
Liquidator Contract - tests

defistar:
- LiquidatorHarness contract for tests
- Tests for liquidator functions

### DIFF
--- a/tests/Contracts/ComptrollerHarness.sol
+++ b/tests/Contracts/ComptrollerHarness.sol
@@ -195,7 +195,7 @@ contract BoolComptroller is ComptrollerInterface {
     bool verifyLiquidateBorrow = true;
     bool verifySeize = true;
     bool verifyTransfer = true;
-
+    uint public liquidationIncentiveMantissa = 11e17;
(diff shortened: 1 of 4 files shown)


kkirka:
Ok, so here I see 6 more or less independent contributions:
1) Liquidator contract fixes (that are already in the target branch)
2) Comptroller harness change, that makes it possible to get and set liquidation incentive mantissa
3) vToken/vToken liquidation test
4) vBNB/vToken liquidation test
5) vBNB/vBNB liquidation test
6) LiquidatorHarness contract and the corresponding tests

Ideally, the commit history presented for review should be equivalent to the "independent contributions" – the reviewers shouldn't care how exactly you fixed problems during development (if it's not a problem pointed out to during the review – in this case it's interesting to know how the code evolved).

I propose we make the following changes to the history:
1) Split commit `554cae1` into 3 atomic commits=
2) Reorder the commits to form 6 groups
3) Remove those commits that touch the Liquidator (these changes are already integrated in feature/liquidator)
4) Use `fixup` to join the groups

PR #98, opened 2022-01-20
VRT Converter and XVS Vesting Contracts

defistar:
## Description

Users can Convert their VRT tokens with a daily Limit
VRT converted to XVS is Vested for a fixed Period of time
Vested XVS can be withdrawn by user

## Checklist


kkirka (commented):
I added some stylistic nitpicks here, but I did not check the vesting formulas in detail. We will have to reconsider our approach to vesting and daily limits, taking the community feedback into account – let's discuss the required changes in more detail today.

PR #130, opened 2022-06-13
Fine grained pause control

kkirka:
## Description

**Problem:** In case of any incident, we would like to be able to just pause certain actions rather than the whole protocol.

**Solution:** Add _setActionsPaused function that accepts a list of markets and actions to pause on these markets. Refactor the current approach to pausing to save some Comptroller space.

**Note:** This change is a bit controversial since it does allow pause guardians to temporarily disable repayBorrow and redeem. This might be worrisome for the community, so I propose we discuss this change publicly before upgrading.

## Checklist

### DIFF
--- a/contracts/Comptroller.sol
+++ b/contracts/Comptroller.sol
@@ -1184,6 +1184,10 @@ contract Comptroller is ComptrollerV9Storage, ComptrollerInterfaceG2, Comptrolle
      * @param paused The new paused state (true=paused, false=unpaused)
      */
     function setActionPausedInternal(address market, Action action, bool paused) internal {
+        require(
+            markets[market].isListed,
+            "cannot pause a market that is not listed"
+        );
         _actionPaused[market][uint(action)] = paused;
         emit ActionPausedMarket(VToken(market), action, paused);
     }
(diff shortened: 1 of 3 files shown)


contracts/Comptroller.sol
```diff
@@ -112,9 +112,14 @@ contract Comptroller is ComptrollerV7Storage, ComptrollerInterfaceG2, Comptrolle
         admin = msg.sender;
     }
 
-    modifier onlyProtocolAllowed {
```
0xlucian, line 115: what is the reasoning behind changing this modifier into a function ?

kkirka: Comptroller size. The current comptroller in the core pool is too close to the size limit, and modifiers occupy lots of space because they are copied verbatim into every function (instead of a bare JUMP for private method calls)

defcon022 (approved):
LGTM

contracts/Comptroller.sol
0xlucian: Thanks

0xlucian:
LGTM as well

0xlucian (approved):
lgtm

PR #139, opened 2022-08-08
[VEN-474]Fix: Reentrancy issue.

defcon022:
## Description

This PR contains the Re-entrancy bugs fixes in the vToken contract. In depth detail of this issue is explained in this [doc](https://docs.google.com/document/d/1JCjOeeV7mifn3CS5bMra9S9jgMPK84enkEFUQ1Ke58o/edit#heading=h.noqquoj45acx).

## Checklist


defcon022:
Have to write test cases for the fix. since we are migrating to Hardhat, will write the test cases after the Migration.

0xlucian (approved):

tests/Utils/Venus.js
```diff
@@ -181,7 +181,7 @@ async function makeVToken(opts = {}) {
       vToken = await saddle.getContractAt("VXvsLikeDelegate", vDelegator._address);
       break;
 
-    case "vbep20":
+    case 'vbep20':
```
coreyar, line 184: nit: At some point it would be nice to write this in typescript

tests/hardhat/EvilXToken.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    const user = account1;
+    const cf1 = 0.5,
+      cf2 = 0.666,
+      cf3 = 0,
+      up1 = 3,
+      up2 = 2.718,
+      up3 = 1;
+
+    const comptrollerLensFactory = await ethers.getContractFactory("ComptrollerLens");
```
coreyar, line 26: We should have deployment scripts that deploy contracts for tests instead of manually deploying them in the test.

Hardhat-deploy will do that when you run `hardhat test`. I am curious though how you would run tests withouy having state dependent on other tests

tests/Utils/Venus.js
defcon022: This does not include in this scope.

tests/hardhat/EvilXToken.ts
defcon022: 1. This is not in the scope of this ticket. 
2. The deployment script for the core pool has not been made till now thats why we are using hardhat upgrade to deploy the tests.
3. Upgradation of these contracts will be done manually.
cc: @kkirka

tests/Utils/Venus.js
coreyar: yes this is why I have marked the comment as a nit

tests/hardhat/EvilXToken.ts
coreyar: I agree refactoring to use deploy scripts is out of scope of this ticket. What brought this to my attention is the really long function. This is a code smell and can be difficult to maintain. In reading through our tests I notice lots of duplicate code. Writing new tests so that they are more maintainable should be our goal. 

The easiest to improve this for now would be to break it out into functions that can be used in this test and others.

Another idea on how we could improve this is to move all the deployments that we expect `hardhat-deploy` to replace into a separate function. That way when we do write the deployment scripts it is clear how to refactor this test.

coreyar: or an even simpler refactoring that improves on the way this test is currently written would be to remove setup (deployments) to a before hook

coreyar:
> Have to write test cases for the fix. since we are migrating to Hardhat, will update the test cases after the Migration.

This should be complete. In the package.json hardhat is used to run tests. I think what is left is migrating some existing tests to hardhat

defcon022:
> > Have to write test cases for the fix. since we are migrating to Hardhat, will update the test cases after the Migration.
> 
> This should be complete. In the package.json hardhat is used to run tests. I think what is left is migrating some existing tests to hardhat.

Test case for this fix is already migrated to hardhat.

tests/hardhat/EvilXToken.ts
defcon022: Yes, we should use deployments in the before hook if tests need to deploy the same contracts several time or for multiple tests, but here contracts are deployed one time only thats why I did not use before hook.

contracts/VToken.sol
```diff
@@ -774,6 +774,10 @@ contract VToken is VTokenInterface, Exponential, TokenErrorReporter {
         // EFFECTS & INTERACTIONS
         // (No safe failures beyond this point)
 
+        /* We write previously calculated values into storage */
+        totalSupply = vars.totalSupplyNew;
+        accountTokens[redeemer] = vars.accountTokensNew;
```
kkirka, line 779: If we move state changes here, we can't `return failOpaque(...)` anymore, we can only revert. Can you please replace all the error conditions to reverts after the state changes are made?

tests/hardhat/EvilXToken.ts
kkirka: I think the main reason for the argument here is that this test looks a lot like an integrational test: it deploys multiple contracts and only makes use of a limited number of fakes. This resembles the old "scenario" tests.

I agree that such complex tests should use deployment scripts. I don't have any opinion wrt this ticket vs other ticket, I'm fine with either.

I don't think unit tests should use the deployment scripts though. In unit tests we're interested in the behavior of a particular contract in isolation: whether it indeed makes external calls, whether the state of this contract is as expected, whether this contract emits the necessary events, etc. For such tests, we'd need a specific deployment – usually the tested contract would be a `smock.mock<...>`, and most of the other contracts would be `smock.fake<...>` (with an exception of ERC-20 tokens and maybe comptroller lens when testing comptroller as it is an irrevokable part of the comptroller itself).

kkirka (changes requested):

tests/Utils/Venus.js
coreyar: I think this is also an unintended change. It should be double quotes

tests/hardhat/EvilXToken.ts
coreyar: yeah that makes a lot of sense that unit tests would use the hardhat deploy scripts. I still think we should have a pattern that reduces the size of test cases so that they are easier to read and maintain. I think generally if different tests have different deployments they would be in different suites so the before hook is a safe place to deploy contracts

contracts/VToken.sol
Manmeet2018: But I don't get this. As the state changes are just moved above the transfer and there is no check of state anywhere down. so it will not effect the case.

tests/hardhat/EvilXToken.ts
Manmeet2018: ok will add it.

contracts/VToken.sol
Manmeet2018: @kkirka any update?

defcon022: But I don't get this. As the state changes are just moved above the transfer and there is no check of state anywhere down. so it will not effect the case.

tests/hardhat/EvilXToken.ts
defcon022: added.

coreyar (approved):

contracts/Tokens/VTokens/VToken.sol
```diff
@@ -708,7 +708,8 @@ contract VToken is VTokenInterface, Exponential, TokenErrorReporter {
         /* exchangeRate = invoke Exchange Rate Stored() */
         (vars.mathErr, vars.exchangeRateMantissa) = exchangeRateStoredInternal();
         if (vars.mathErr != MathError.NO_ERROR) {
-            return failOpaque(Error.MATH_ERROR, FailureInfo.REDEEM_EXCHANGE_RATE_READ_FAILED, uint(vars.mathErr));
+            failOpaque(Error.MATH_ERROR, FailureInfo.REDEEM_EXCHANGE_RATE_READ_FAILED, uint(vars.mathErr));
+            revert();
```
kkirka, line 712: We can't get events from failed txs AFAIK, so I propose just we just do this for now:

```suggestion
            revert("math error");
```

In the long run, we should upgrade the compiler and use custom errors IMO

defcon022: Still we would get events there because I did not remove `failOpaque(Error.MATH_ERROR, FailureInfo.REDEEM_EXCHANGE_RATE_READ_FAILED, uint(vars.mathErr));` @kkirka @coreyar

coreyar: My understanding is that events are not emitted if a transaction is reverted. There seem to be two tests for this - `returns error from redeemFresh without emitting any extra log`

coreyar: In this case we can remove the `failOpaque` calls when using revert

narayanprusty (approved):

kkirka (approved):

PR #143, opened 2022-08-16
Replace the existing tooling with hardhat-based one

0xlucian:
This PR removes the original compound dependencies and brings in a modern hardhat version and some plugins.

This change expectedly breaks the existing scripts and tests.


0xlucian (commented):
LGTM @kkirka @defcon022 we need some reviews, since I opened the PR and cannot approve my own PRs

defcon022 (approved):
LGTM :)

kkirka (approved):
Approving on @0xlucian's behalf, let's merge

PR #144, opened 2022-08-18
Migrate comptroller unit tests to hardhat

kkirka:
This PR is the first step of migrating the tests to smock/hardhat-based framework. Please share your comments & suggestions, especially on how to improve readability of fixtures.


0xlucian:
IMO in the context of smock fixtures are implemented as clean as they could. Usually fixtures are used with utilizing the deploy scripts code (hardhat-deploy plugin has made it very easy), thus making the tests more readable. I propose we continue the unit tests with this approach of fixtures (if no one comes with a cleaner idea). When it comes to migrating scenarios, I propose we create deploy scripts with and then migrate the scenarios in hardhat tests using hardhat-deploy plugin fixtures.

RoseRompkxm (approved):

PR #146, opened 2022-08-23
Fix hardhat config & tests

kkirka:
This PR is a continuation of our work on migrating to hardhat. It adds typechain mappings generation to hardhat config, adds the missing imported file and updates the Dockerfile to run hardhat compilation instead of saddle compilation.


0xlucian (approved):
Lgtm

PR #147, opened 2022-08-28
Use Memory Type Variable Instead of Storage Type Variable in Event to Save Gas 

HighBe:
Hi, we are a research group on programming languages and software engineering. We recently have conducted a systematic study about Solidity event usage, evolution, and impact, and we are attempting to build a tool to improve the practice of Solidity event use based on our findings. We have tried our prototype tool on some of the most popular GitHub Solidity repositories, and for your repository, we find a potential optimization of gas consumption arisen from event use.

The point is that when we use emit operation to store the value of a certain variable, `local memory type variable` would be preferable to `storage type (state) variable` if they hold the same value. The reason is that an extra SLOAD operation would be needed to access the variable if it is storage type, and the SLOAD operation costs 800 gas.

For your repository, we find that several event uses can be improved.

### DIFF
--- a/contracts/JumpRateModel.sol
+++ b/contracts/JumpRateModel.sol
@@ -50,7 +50,7 @@ contract JumpRateModel is InterestRateModel {
         jumpMultiplierPerBlock = jumpMultiplierPerYear.div(blocksPerYear);
         kink = kink_;
 
-        emit NewInterestParams(baseRatePerBlock, multiplierPerBlock, jumpMultiplierPerBlock, kink);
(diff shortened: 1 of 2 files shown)


0xlucian:
seems logical, will get back with a review in a while, since we are out of resources for now

coreyar:
@HighBe We'd like to make sure you get credit for contributing to Venus. In order to do that you need to resolve the conflicts. Apologies for taking so long to address this PR. We were heads down in feature work. Thanks again for taking the time to identify this optimization.

coreyar:
This closes issue #145

PR #156, opened 2022-09-16
Optimize XVS claims and blacklist XVS claimers

kkirka:
See the commit messages for details

### DIFF
--- a/contracts/Comptroller.sol
+++ b/contracts/Comptroller.sol
@@ -1256,14 +1256,6 @@ contract Comptroller is ComptrollerV7Storage, ComptrollerInterfaceG2, Comptrolle
      */
     function claimVenus(address[] memory holders, VToken[] memory vTokens, bool borrowers, bool suppliers, bool collateral) public {
         uint j;
-        // Save shortfalls of all holders
-        // if there is a positive shortfall, the XVS reward is accrued,
-        // but won't be granted to this holder
-        uint[] memory shortfalls = new uint[](holders.length);
-        for (j = 0; j < holders.length; j++) {
-            (, , uint shortfall) = getHypotheticalAccountLiquidityInternal(holders[j], VToken(0), 0, 0);
-            shortfalls[j] = shortfall;
-        }
         for (uint i = 0; i < vTokens.length; i++) {
             VToken vToken = vTokens[i];
             require(markets[address(vToken)].isListed, "not listed market");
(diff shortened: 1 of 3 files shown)


contracts/Comptroller.sol
```diff
@@ -1305,6 +1304,15 @@ contract Comptroller is ComptrollerV7Storage, ComptrollerInterfaceG2, Comptrolle
      * @return The amount of XVS which was NOT transferred to the user
      */
     function grantXVSInternal(address user, uint amount, uint shortfall, bool collateral) internal returns (uint) {
+        // If the user is blacklisted, they can't get XVS rewards
+        require(
+            user != 0xEF044206Db68E40520BfA82D45419d498b4bc7Bf
```
stubbi, line 1309: my 2 cents: move those addresses to some "blacklisted" array and use some "user not in blacklisted" if that's possible to make the code slightly simpler and more readable?

stubbi: if that is even how this piece of code works

defistar (approved):

defcon022 (approved):

contracts/Comptroller.sol
kkirka: Yeah, I thought about this, but it will increase gas consumption, and it isn't much more readable IMO:
```
address[] memory blacklisted = [
    0xEF044206Db68E40520BfA82D45419d498b4bc7Bf,
    0x7589dD3355DAE848FDbF75044A3495351655cB1A,
    0x33df7a7F6D44307E1e5F3B15975b47515e5524c0,
    0x24e77E5b74B30b026E9996e4bc3329c881e24968
];

for (uint k = 0; k < blacklisted.length; ++k) {
    require(user != blacklisted[k], "Blacklisted");
}
```

A better solution would be to have a mapping in storage, like
```
mapping (address => bool) blacklistedUsers;

function setBlacklisted(address user, bool blacklisted) external { ... }

function grantXVSInternal(...) {
    require(!blacklistedUsers[user], "Blacklisted");
    ...
}
```

The reason I haven't done it this way is that:
1) we will likely hit the upper bound on contract size since we introduce a new `external` function
2) the storage layout will conflict with what we have in our `develop` branch currently, which may lead to some hard to catch errors during the upgrade

But this is debatable ofc

stubbi: oh, I see. Yes, that makes lots of sense and sounds reasonable. Since I haven't done any sc development myself before, I am not aware of all the implications and it sounds to me like a well thought through solution 👍

coreyar (approved):

PR #157, opened 2022-10-03
XVSStakingLens - get Staked XVS Details of an Account

defistar:
## Description

1. Get Staked XVSAmount of a user
2. Get PendingWithdrawal amount - from Staked XVS  of a user


kkirka (approved):

contracts/Lens/XVSStakingLens.sol
```diff
@@ -0,0 +1,32 @@
+pragma solidity ^0.5.16;
```
coreyar, line 1: When adding new contracts should we also add the deployment script for them?

0xlucian (approved):

PR #160, opened 2022-10-07
Optimize comptroller size yet again

kkirka:
This PR reverts in case access controls fail, and hides some deprecated storage fields to save space on public getters


0xlucian (approved):
LGTM, good job with removing silent reverts

PR #166, opened 2022-10-21
[VEN-467] VAI Stability Fee

narayanprusty:
## Description

Resolves #

## Checklist


defistar (approved):

contracts/VAIController.sol
```diff
@@ -410,10 +437,175 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+        accrualBlockNumber = getBlockNumber();
+    }
+        
+    /**
+     * @dev Set VAI borrow base rate
+     */
+    function _setBaseRate(uint newBaseRateMantissa) external returns (uint) {
+        // Check caller is admin
+        if (msg.sender != admin) {
+            return fail(Error.UNAUTHORIZED, FailureInfo.SET_COMPTROLLER_OWNER_CHECK);
```
kkirka, line 568: `SET_COMPTROLLER_OWNER_CHECK` looks weird here, maybe we could just revert?

contracts/VAIController.sol
```diff
@@ -410,10 +437,175 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+        emit NewVAIBaseRate(old, baseRateMantissa);
+    }
+
+    /**
+     * @dev Set VAI borrow float rate
+     */
+    function _setFloatRate(uint newFloatRateMantissa) external returns (uint) {
+        // Check caller is admin
+        if (msg.sender != admin) {
+            return fail(Error.UNAUTHORIZED, FailureInfo.SET_COMPTROLLER_OWNER_CHECK);
```
kkirka, line 581: Here and in `_setReceiver` as well

contracts/VAIController.sol
```diff
@@ -158,21 +178,23 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
      * @return (uint, uint) An error code (0=success, otherwise a failure, see ErrorReporter.sol), and the actual repayment amount.
      */
     function repayVAIFresh(address payer, address borrower, uint repayAmount) internal returns (uint, uint) {
+        accrueVAIInterest();
```
kkirka, line 181: `Fresh` usually means the interest has been accrued already. Maybe move interest accrual to `repayVAI`?

contracts/Comptroller.sol
```diff
@@ -10,12 +10,13 @@ import "./Governance/XVS.sol";
...
 import "./ComptrollerLensInterface.sol";
 import "./IAccessControlManager.sol";
+import "./CarefulMath.sol";
 
 /**
  * @title Venus's Comptroller Contract
  * @author Venus
  */
-contract Comptroller is ComptrollerV9Storage, ComptrollerInterfaceG2, ComptrollerErrorReporter, ExponentialNoError {
+contract Comptroller is ComptrollerV9Storage, ComptrollerInterfaceG2, ComptrollerErrorReporter, ExponentialNoError, CarefulMath {
```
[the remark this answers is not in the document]
narayanprusty: Yes we don't need it. I forgot to remove it.

contracts/VAIController.sol
```diff
@@ -410,10 +437,175 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+            } else {
+                return baseRateMantissa;
+            }
+        } else {
+            return 0;
+        }
+    }
+
+    function getVAIRepayRatePerBlock() public view returns (uint) {
+        PriceOracle oracle = ComptrollerImplInterface(address(comptroller)).oracle();
```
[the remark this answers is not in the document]
narayanprusty: No we don't need to. This also I forgot to remove.

contracts/VAIController.sol
```diff
@@ -93,7 +104,14 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
 
-            (vars.mathErr, vars.accountMintVAINew) = addUInt(ComptrollerImplInterface(address(comptroller)).mintedVAIs(minter), mintVAIAmount);
+            // Calculate the minted balance based on interest index
+            uint totalMintedVAI = ComptrollerImplInterface(address(comptroller)).mintedVAIs(minter);
+
+            if (totalMintedVAI > 0) {
+                totalMintedVAI = getVAIRepayAmount(minter);
+            }  
+
+            (vars.mathErr, vars.accountMintVAINew) = addUInt(totalMintedVAI, mintVAIAmount);
```
kkirka, line 114: So here we put interest into the principal which seems ok, and this is exactly what we do for VTokens. However, later we treat interest and principal separately to compute the percentages to burn and to put to treasury. This means that the following two sequences of actions will lead to different results:

```
Sequence A:
1. mintVAI(100 VAI)
2. wait for 1 year
3. mint(0 VAI)
4. repayVAI(MAX_UINT256)
```

```
Sequence B:
1. mintVAI(100 VAI)
2. wait for 1 year
3. repayVAI(MAX_UINT256)
```

In sequence A the entire interest will be burned, while in sequence B the interest will go into treasury.

kkirka (commented):
I left a comment, and I'm not entirely sure what to do with this. I would like to confirm I understood the math correctly, can you please confirm it? If my understanding is correct, let's discuss what we can do about it.

contracts/VAIController.sol
narayanprusty: Yes I think you understood it correctly. Should we introduce another variable to track user's interest? And transfer to treasury whatever is in that variable.

narayanprusty: > This looks like a right thing to do, but I'm not sure what formulas we should use in this case. Does this look right?
> 
> During minting we save interest to a separate variable `pastInterest` (mapping from address to uint): pastInterest=mintedVAI∗(globalIndex/minterIndex) mintedVAI+=mintAmount minterIndex=globalIndex
> 
> Then to compute the owed amount and the proportion that would go to treasury/burn we'd use: owedAmount=(mintedVAI+pastInterest)∗(globalIndex/minterIndex) principal=mintedVAI (goes to burn) totalInterest=owedAmount−principal (goes to treasury)
> 
> What would be the formula for partial repayment? Will we subtract anything from pastInterest?

Yes I think in case of partial repayment we don't need to subtract anything from pastInterest.

narayanprusty: > Other options should be possible as well. Do we really need to track "treasury amount" per user? Maybe we can simplify the formulas by using global market-level math.

Do you have any formulas in mind? I cannot think of any.

kkirka: > Yes I think in case of partial repayment we don't need to subtract anything from pastInterest.

This makes the conditions kinda complex:
if repayAmount < mintedVAI, repay a part of the principal
else if repayAmount < mintedVAI + current interest, repay a part of the principal and a part of the current interest,
else if repayAmount < mintedVAI + past interest + current interest, **???**
otherwise, repay everything

I'd rather simplify this if possible, but I can't think of a formula from the top of my head. Something like track the total amount of minted VAIs (without interest) and total accumulated interest, compute a percentage based on these values, put this percentage to treasury and burn the rest? 🤔

narayanprusty: I think we can simplify the above by keeping mintedVAI as it is currently and just use totalInterest to track the amount of mintedVAI that's actually the interest amount. 

```
if repayAmount < mintedVAI
  mintedVAI = mintedVAI - repayAmount
else 
  mintedVAI = 0


if totalInterest > mintedVAI
  treasuryTransferAmount = totalInterest - mintedVAI
  totalInterest = mintedVAI
else 
  treasuryTransferAmount = totalInterest
  totalInterest = 0
```

Also we can include the currentInterest in the totalInterest before running the above formula.

narayanprusty: As discussed we will have `mintedVAI` as the total VAI minted by user and `totalInterest` representing interest accrued by the user

narayanprusty: @kkirka I have updated the implementation. Now we have a variable pastInterestOfUser tracking the amount of mintedVaiOfUser that represents the user's accrued interest. During repayment I use this formulas to update mintedVAI and pastInterest:

```
repayAmount = mintedVaiOfUser * (globalIndex / userMintIndex)
partOfRepayAmountUserIsRepaying = amountUserIsRepaying / repayAmount
totalCurrentInterestOfUser =repayAmount - mintedVaiOfUser +pastInterestOfUser 
partOfCurrentInterestOfUser =totalCurrentInterestOfUser *partOfRepayAmountUserIsRepaying 
treasuryAmount = treasuryAmount + partOfCurrentInterestOfUser
burn = (repayAmount - totalCurrentInterestOfUser) * partOfRepayAmountUserIsRepaying  
partOfPastInterestOfIser =pastInterestOfUser *partOfRepayAmountUserIsRepaying   
mintedVaiOfUser = mintedVaiOfUser - burn  -partOfPastInterestOfIser 
pastInterestOfUser=pastInterestOfUser -(pastInterestOfUser *partOfRepayAmountUserIsRepaying ) 
```

contracts/VAIController.sol
```diff
@@ -7,6 +7,10 @@ import "./Exponential.sol";
 import "./VAIControllerStorage.sol";
 import "./VAIUnitroller.sol";
 import "./VAI/VAI.sol";
+import "./ComptrollerStorage.sol";
+import "./Comptroller.sol";
+
+import "hardhat/console.sol";
```
defcon022, line 13: Remove `import "hardhat/console.sol";`
```suggestion
```

contracts/VAIController.sol
```diff
@@ -64,6 +68,18 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
      */
     event MintFee(address minter, uint feeAmount);
 
+    /// @notice Emiitted when VAI base rate is changed
+    event NewVAIBaseRate(uint oldBaseRateMantissa, uint newBaseRateMantissa);
```
defcon022, line 72: Can we use `uint256` instead of `uint`?
```suggestion
    event NewVAIBaseRate(uint256 oldBaseRateMantissa, uint256 newBaseRateMantissa);
```

PR #170, opened 2022-11-14
VAI Minting

narayanprusty:
## Description

1. Mint VAI based on weighted supply
2. VAI Mint Cap

Resolves #

## Checklist

### DIFF
--- a/contracts/VAIController.sol
+++ b/contracts/VAIController.sol
@@ -368,6 +368,7 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
         uint oErr;
         MathError mErr;
         uint sumSupply;
+        uint marketSupply;
         uint sumBorrowPlusEffects;
         uint vTokenBalance;
         uint borrowBalance;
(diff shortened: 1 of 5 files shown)


contracts/VAIController.sol
```diff
@@ -95,6 +101,12 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
 
             address minter = msg.sender;
 
+            (vars.mathErr, totalMintedVAI) = addUInt(totalMintedVAI, mintVAIAmount);
+
+            if (mintCap != 0) {
```
coreyar, line 106: do we need to handle a case where the mintCap is 0? We could throw the same error in this case too

contracts/VAIController.sol
```diff
@@ -402,6 +416,17 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
                 return (uint(Error.MATH_ERROR), 0);
             }
 
+            (, uint collateralFactorMantissa,) = Comptroller(address(comptroller)).markets(address(enteredMarkets[i]));
+            (vars.mErr, vars.sumSupply) = mulUInt(vars.sumSupply, collateralFactorMantissa);
+            if (vars.mErr != MathError.NO_ERROR) {
+                return (uint(Error.MATH_ERROR), 0);
```
coreyar, line 422: Instead of returning new error codes should we instead add requires to slowly migrate us off of returning errors?

contracts/VAIController.sol
```diff
@@ -402,6 +416,17 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
                 return (uint(Error.MATH_ERROR), 0);
             }
 
+            (, uint collateralFactorMantissa,) = Comptroller(address(comptroller)).markets(address(enteredMarkets[i]));
+            (vars.mErr, vars.sumSupply) = mulUInt(vars.sumSupply, collateralFactorMantissa);
```
kkirka, line 420: Wait, I think you should only include `enteredMarkets[i]` only here, but now you do sth like

Market2.collateralFactor * (Market1.collateralFactor * supply1 + supply2))

but it should be

Market1.collateralFactor * supply1 + Market2.collateralFactor * supply2

narayanprusty: @kkirka Fixed it. Now I am taking weighted supply of each market and summing them up to find the total weighted supply.

contracts/VAIController.sol
0xlucian: Agree, we should try to move off from "silent" errors. They make writing integration tests slower than in we just revert with error

contracts/VAIController.sol
narayanprusty: @coreyar In vToken borrow caps we are using 0 for infinite borrow. But as discussed in DSU, I will changed it to such that 0 mean no minting is allowed and by default infinite minting is enabled.

contracts/VAIController.sol
narayanprusty: As discussed, we will resolve these in a separate PR.

contracts/VAIController.sol
narayanprusty: @kkirka I agree. I have updated the logic.

kkirka (approved):

PR #177, opened 2022-11-28
Add edit nat spec

coreyar:
## Description

Adds nat spec for VenusLens and edits natspec for governorBravo

### DIFF
--- a/contracts/Governance/GovernorBravoDelegate.sol
+++ b/contracts/Governance/GovernorBravoDelegate.sol
@@ -38,9 +38,8 @@ contract GovernorBravoDelegate is GovernorBravoDelegateStorageV2, GovernorBravoE
     /**
       * @notice Used to initialize the contract during delegator contructor
       * @param xvsVault_ The address of the XvsVault
-      * @param proposalConfigs_ Governance confifgs for each GovernanceType
-      * @param timelocks Timelock addresses for each GovernanceType
-      * @param guardian_ address of governance guardian
+      * @param proposalConfigs_ Governance configs for each governance route
+      * @param timelocks Timelock addresses for each governance route
       */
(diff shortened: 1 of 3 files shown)


contracts/Governance/GovernorBravoDelegate.sol
```diff
@@ -193,7 +193,7 @@ contract GovernorBravoDelegate is GovernorBravoDelegateStorageV2, GovernorBravoE
     /**
       * @notice Gets actions of a proposal
       * @param proposalId the id of the proposal
-      * @return Targets, values, signatures, and calldatas of the proposal actions
+      * @return targets, values, signatures, and calldatas of the proposal actions
```
0xlucian, line 196: the auto generation of docs may fail fail because it is not case sensitive and it expects after @return to be the names of the return variables (if the return names are defined in function signatures)

0xlucian (changes requested):
please verify that all natspec return variables are with lowercase since docgen tool will throw errors

contracts/Governance/GovernorBravoDelegate.sol
coreyar: The params are case sensitive because they are named. Return types in the function signature don't define a variable name for the return because a return may not return assignment to a variable. It could return a value returned directly from another function call so there isn't a variable to match. The return in the natspec doc is prose while the params are a table which is why the first word has to match

This is a good catch though, there were misspelled params that I've corrected.

coreyar:
> please verify that all natspec return variables are with lowercase since docgen tool will throw errors

Done. I added a job to check this

0xlucian (approved):

defistar (approved):

PR #184, opened 2022-12-08
[VEN-792] Stop XVS Reward during Cooldown

narayanprusty:
## Description

- Added a new state variable `totalPendingWithdrawals` to track pending withdrawals across all users. 
- We calculate and transfer rewards in requestWithdrawal instead executeWithdrawal 
- If a user has unexecuted pending withdrawal after upgrade then then need to execute that before requesting new withdrawals
- Added `afterUpgrade` flag for WithdrawalRequest so that we can handle existing withdrawals correctly and pay rewards in executeWithdrawal.

Resolves #VEN-792

## Checklist

### DIFF
--- a/contracts/XVSVault/XVSVault.sol
+++ b/contracts/XVSVault/XVSVault.sol
@@ -340,15 +340,45 @@ contract XVSVault is XVSVaultStorage, ECDSA {
             user.amount = user.amount.sub(beforeUpgradeWithdrawalAmount);
             user.rewardDebt = user.amount.mul(pool.accRewardPerShare).div(1e12);
             pool.token.safeTransfer(address(msg.sender), beforeUpgradeWithdrawalAmount);
-        }
-        
-        if (afterUpgradeWithdrawalAmount > 0) {
+        } else {
             user.amount = user.amount.sub(afterUpgradeWithdrawalAmount);
             totalPendingWithdrawals[_rewardToken][_pid] = totalPendingWithdrawals[_rewardToken][_pid].sub(afterUpgradeWithdrawalAmount);
             pool.token.safeTransfer(address(msg.sender), afterUpgradeWithdrawalAmount);
         }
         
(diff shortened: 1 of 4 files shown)


contracts/XVSVault/XVSVault.sol
```diff
@@ -348,10 +343,20 @@ contract XVSVault is XVSVaultStorage, ECDSA {
         require(user.amount >= user.pendingWithdrawals.add(_amount), "requested amount is invalid");
 
         PoolInfo storage pool = poolInfos[_rewardToken][_pid];
+
+        _updatePool(_rewardToken, _pid);
```
coreyar, line 347: Nice. So if I understand correctly this was moved to follow the checks-effects-interaction pattern?

tests/hardhat/XVS/XVSVault.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+describe("XVSVault", async () => {
+  let user1: Wallet;
+  let xvs: XVS;
+  let xvsVault: XVSVault;
+  let xvsStore: XVSStore;
+
+  before("get signers", async () => {
+    [user1] = await (ethers as any).getSigners();
```
coreyar, line 27: nit: we typically call the first signer the root/ deployer.  Could we use getNamedSigners here?

I haven't finished thinking this through but it seems like we could setup different named accounts to use for different scenarios tests

coreyar (approved):

contracts/XVSVault/XVSVault.sol
narayanprusty: We moved it because now we are going to pay rewards in `requestWithdrawal` so for that `_updatePool` needs to be called before calculating rewards.

kkirka (approved):

tests/hardhat/XVS/XVSVault.ts
narayanprusty: I have updated the tests to use first account as `deployer` and second account for `user`

PR #193, opened 2022-12-27
fix: set initial stability fee index for past minters

kkirka:
The stability fee implementation proposed in [VIP-80](https://app.venus.io/governance/proposal/80) contained a bug leading to incorrect treatment of existing VAI minters. This PR sets initial minter index for those who minted VAI before the stability fee was introduced.


defcon022 (approved):
LGTM :)

PR #192, opened 2022-12-23
GovernanceV4- routes simulation

0xlucian:
## Description

This PR adds forked mainnet tests for the new GovernanceV4 upgrade. Along with the tests, it introduces a new approach of creating forked mainnet tests by calling `setForkBlock(blockNumber);` in order to fork a chain from a certain block. This allows us to specify the fork block more glanularly and is a bit more flexible.

## Checklist

- [ ] I have updated the documentation to account for the changes in the code.
- [ ] If I added new functionality, I added tests covering it.
- [ ] If I fixed a bug, I added a test preventing this bug from silently reappearing again.
- [ ] My contribution follows [Venus contribution guidelines](docs/CONTRIBUTING.md).

PR #194, opened 2022-12-27
test: add simulations for VAI upgrade and post-upgrade behavior

kkirka:
This PR builds on top of https://github.com/VenusProtocol/venus-protocol/pull/192 and brings in utility functions for simulating VIPs, and a simulation for VAI stability fee pugrade


script/hardhat/fork/govUpdateTest.ts
```diff
@@ -0,0 +1,325 @@
...
+import { expect } from "chai";
+import { parseUnits } from "ethers/lib/utils";
+import { ethers } from "hardhat";
+
+import { Comptroller, GovernorBravoDelegate, IAccessControlManager } from "../../../typechain";
+import { TimelockInterface } from "../../../typechain/contracts/Governance/GovernorAlpha2.sol";
+import { getCalldatas, setForkBlock } from "./vip-framework/utils";
+
+const ONE_HOUR = 3600; // 1 hour
+const SIX_HOURS = 6 * 3600; // 6 hours
```
coreyar, line 12: nit: could multiply the constant, which will be nice in the event the number of seconds in an hour changes 😝 
```suggestion
const SIX_HOURS = 6 * ONE_HOUR; // 6 hours
```

script/hardhat/fork/govUpdateTest.ts
```diff
@@ -0,0 +1,325 @@
...
+let timeLockFastTrack: TimelockInterface;
+let timeLockCritical: TimelockInterface;
+
+const initMainnetUser = async (user: string, balance: number) => {
+  await impersonateAccount(user);
+  await setBalance(user, balance);
+  return ethers.getSigner(user);
+};
+const governanceFixture = async (): Promise<void> => {
+  //Mandatory when creating forked tests. One must specify form which block the fork should be
```
coreyar, line 68:
```suggestion
  // Mandatory when creating forked tests. One must specify form which block the fork should be
```

chechu (commented):
Awesome job! Would it make sense to move the vip-framework to an isolated package (maybe in the future)? I was wondering if we could use it in the future for VIP's related to isolated pools

kkirka:
> Awesome job! Would it make sense to move the vip-framework to an isolated package (maybe in the future)? I was wondering if we could use it in the future for VIP's related to isolated pools

I'd rather move the whole governance into its own package. Governance is something that is relevant not only to core, but also to isolated pools and oracle, and maybe other contracts we will have in future.

chechu (approved):

chechu:
> I'd rather move the whole governance into its own package. Governance is something that is relevant not only to core, but also to isolated pools and oracle, and maybe other contracts we will have in future.
> 
> This would also allow us to version governance independently from core/IL/oracle.

https://jira.toolsfdg.net/browse/VEN-930

PR #195, opened 2022-12-27
chore: set liquidator compiler version to more stable

kkirka:
Auditors often report that using too recent compiler versions may lead to facing yet undiscovered bugs. Version 0.8.13 is stable, and the known issues there are not relevant to our contracts. We use 0.8.13 in our other repos, so in addition to security reasons, I'm also making this change for consistency.


chechu (approved):
Will we have to redeploy this contract in the main net, considering this change?

kkirka:
> Will we have to redeploy this contract in the main net, considering this change?

It's already deployed actually

PR #200, opened 2023-01-16
[VEN-747] Edit nat spec and order vtoken contracts

coreyar:
## Description

Adds linting ordering and orders entities
Edits nat spec for clarity and per guidelines

### DIFF
--- a/contracts/Tokens/VTokens/VBNB.sol
+++ b/contracts/Tokens/VTokens/VBNB.sol
@@ -3,13 +3,13 @@ pragma solidity ^0.5.16;
 import "./VToken.sol";
 
 /**
- * @title Venus's VBNB Contract
- * @notice VToken which wraps BNB
+ * @title Venus's vBNB Contract
+ * @notice vToken which wraps BNB
  * @author Venus
  */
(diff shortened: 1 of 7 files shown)


contracts/Tokens/VTokens/VBNB.sol
```diff
@@ -36,12 +36,22 @@ contract VBNB is VToken {
...
+        requireNoError(err, "mint failed");
+    }
+
     /*** User Interface ***/
 
     /**
      * @notice Sender supplies assets into the market and receives vTokens in exchange
      * @dev Reverts upon any failure
      */
+    // @custom:events Emits Transfer event
```
chechu, line 53: We could avoid mixing styles in the same block. I would add the `custom:events` tags into the `/** ... */` block

chechu: I'm sorry, now I read your comment in Slack. I suppose you don't want to add these comments in the final doc, right? BTW, why?

chechu (approved):
LGTM

defcon022 (approved):
LGTM

PR #202, opened 2023-01-25
[VEN-883] refactor pendingVenus()

0xlucian:
## Description

This PR refactors `pendingVenus` function to return more structured data about pendingRewards.

Resolves : VEN-883

## Checklist

### DIFF
--- a/contracts/Lens/VenusLens.sol
+++ b/contracts/Lens/VenusLens.sol
@@ -129,7 +129,6 @@ contract VenusLens is ExponentialNoError {
         PendingReward[] pendingRewards;
     }
 
-
     /**
      * @notice Query the metadata of a vToken by its address
      * @param vToken The address of the vToken to fetch VTokenMetadata
--- a/contracts/Lens/VenusLens.sol
+++ b/contracts/Lens/VenusLens.sol
@@ -606,12 +606,16 @@ contract VenusLens is ExponentialNoError {
      * @param comptroller Address of the comptroller
      * @return Reward object contraining the totalRewards and pending rewards for each market
      */
-    function pendingRewards(address holder, ComptrollerInterface comptroller) external view returns (RewardSummary memory) {
+    function pendingRewards(
+        address holder,
+        ComptrollerInterface comptroller
+    ) external view returns (RewardSummary memory) {
         VToken[] memory vTokens = comptroller.getAllMarkets();
         ClaimVenusLocalVariables memory vars;
         RewardSummary memory rewardSummary;
         rewardSummary.rewardTokenAddress = comptroller.getXVSAddress();
         rewardSummary.totalRewards = comptroller.venusAccrued(holder);
+        rewardSummary.pendingRewards = new PendingReward[](vTokens.length);
         for (uint i = 0; i < vTokens.length; i++) {
             (vars.borrowIndex, vars.borrowBlock) = comptroller.venusBorrowState(address(vTokens[i]));
             VenusMarketState memory borrowState = VenusMarketState({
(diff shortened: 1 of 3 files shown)


contracts/Lens/VenusLens.sol
```diff
@@ -584,14 +584,34 @@ contract VenusLens is ExponentialNoError {
...
+
+    /**
+     * @notice Calculate the total XVS tokens pending and accrued by a user account
      * @param holder Account to query pending XVS
      * @param comptroller Address of the comptroller
-     * @return Total number of accrued XVS that can be claimed
+     * @return Reward object contraining the totalRewards and pending rewards for each market
      */
-    function pendingVenus(address holder, ComptrollerInterface comptroller) external view returns (uint) {
+    function pendingVenus(address holder, ComptrollerInterface comptroller) external view returns (Reward memory) {
```
kkirka, line 609: Maybe we should rename the function as well then?

```suggestion
    function pendingRewards(address holder, ComptrollerInterface comptroller) external view returns (Reward memory) {
```

chechu: Would it make sense to create a new function, keeping the legacy one as it was?

0xlucian: Not really sure if the old function can be used for something, because it gives the same information. Maybe we can leave the same name? @therealemjy  can you give your opinion?
Do we need to have the same interface for both VenusLens and PoolLens?

therealemjy: The old one won't be used anymore once this is deployed, so no need to keep it for legacy purposes.
Regarding the interface, if I understand well both share basically the same interface, except the function from `VenusLens` returns a single `Reward` instead of an array of `Reward`. Then I think what you did there is good.

0xlucian: Funciton name has been changed

chechu (approved):
LGTM

contracts/Lens/VenusLens.sol
```diff
@@ -584,14 +584,38 @@ contract VenusLens is ExponentialNoError {
...
+     */
+    struct PendingReward {
+        address vTokenAddress;
+        uint256 amount;
+    }
+
+    /**
+     * @dev Struct for Reward of a single reward token.
+     */
+    struct RewardSummary {
```
kkirka, line 597: Maybe add distributorAddress = core pool comptroller for compatibility with IL?

contracts/Lens/VenusLens.sol
```diff
@@ -584,14 +584,38 @@ contract VenusLens is ExponentialNoError {
     }
 
     /**
-     * @notice Calculate the total XVS tokens pending or accrued by a user account
+     * @dev Struct for Pending Rewards for per market
+     */
+    struct PendingReward {
```
kkirka, line 589: Let's move structs to the top of the file so that we have less conflicts when we enable linting

https://docs.soliditylang.org/en/v0.8.17/style-guide.html#order-of-layout

contracts/Lens/VenusLens.sol
```diff
@@ -584,14 +584,38 @@ contract VenusLens is ExponentialNoError {
...
+        address holder,
+        ComptrollerInterface comptroller
+    ) external view returns (RewardSummary memory) {
         VToken[] memory vTokens = comptroller.getAllMarkets();
         ClaimVenusLocalVariables memory vars;
+        RewardSummary memory rewardSummary;
+        rewardSummary.rewardTokenAddress = comptroller.getXVSAddress();
+        rewardSummary.totalRewards = comptroller.venusAccrued(holder);
+        rewardSummary.pendingRewards = new PendingReward[](vTokens.length);
         for (uint i = 0; i < vTokens.length; i++) {
```
defcon022, line 619: Gas optimisation and follow everywhere we are using for loop.
* remove re-decleration of uint i to 0
* use ++i instead of i++

contracts/Lens/VenusLens.sol
0xlucian: Done

contracts/Lens/VenusLens.sol
0xlucian: Done.

PR #211, opened 2023-02-17
feat: add delegate borrowing feature

kkirka:
This PR adds delegated borrowing feature. It would allow users to delegate their borrowing power to some account. In `DelegateBorrowers/` directory there is an example delegate – it only allows to swap one debt position to another, e.g. by repaying BUSD and borrowing ETH. The conversion is done based on the current oracle price.

The Comptroller contract is currently oversize by 94 bytes, tests are not yet implemented.

### DIFF
--- a/contracts/DelegateBorrowers/SwapDebtDelegate.sol
+++ b/contracts/DelegateBorrowers/SwapDebtDelegate.sol
@@ -101,10 +101,15 @@ contract SwapDebtDelegate is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable
         uint256 repayAmount
     ) internal returns (uint256 actualRepaymentAmount) {
         IERC20Upgradeable underlying = IERC20Upgradeable(vToken.underlying());
+        uint256 balanceBefore = underlying.balanceOf(address(this));
         underlying.safeTransferFrom(msg.sender, address(this), repayAmount);
+        uint256 balanceAfter = underlying.balanceOf(address(this));
+        uint256 repayAmountMinusFee = balanceAfter - balanceBefore;
 
+        underlying.safeApprove(address(vToken), 0);
+        underlying.safeApprove(address(vToken), repayAmountMinusFee);
         uint256 borrowBalanceBefore = vToken.borrowBalanceCurrent(borrower);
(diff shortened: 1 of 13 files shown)


kkirka:
Initially, I tried creating `addDelegate` and `removeDelegate` functions but it made the Comptroller oversize by 200+ bytes, which isn't acceptable. I merged these into `updateDelegate`, which is a bit less intuitive, but should work 🤷

chechu (approved):

contracts/DelegateBorrowers/SwapDebtDelegate.sol
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * @notice Repays a borrow in repayTo.underlying() and borrows borrowFrom.underlying()
+     * @param borrower The address of the borrower, whose debt to swap
+     * @param repayTo VToken to repay the debt to
+     * @param borrowFrom VToken to borrow from
+     * @param repayAmount The amount to repay in terms of repayTo.underlying()
+     */
+    function swapDebt(
```
narayanprusty, line 57: I think we should emit an event when debt is swapped successfully

contracts/DelegateBorrowers/SwapDebtDelegate.sol
```diff
@@ -0,0 +1,135 @@
...
+        uint256 actualRepaymentAmount = _repay(repayTo, borrower, repayAmount);
+        uint256 amountToBorrow = _convert(repayTo, borrowFrom, actualRepaymentAmount);
+        _borrow(borrowFrom, borrower, amountToBorrow);
+    }
+
+    /**
+     * @notice Transfers tokens, accidentially sent to this contract, to the owner
+     * @param token ERC-20 token to sweep
+     */
+    function sweepTokens(IERC20Upgradeable token) external onlyOwner {
```
narayanprusty, line 72: Here also we should emit an event when tokens are swept successfully

narayanprusty (approved):

defcon022 (approved):

UrAvgDeveloper (approved):

chechu (approved):
LGTM

PR #228, opened 2023-03-16
Swap Audit Fixes

defcon022:
## Description

Resolves VEN-1209

## Checklist


contracts/Swap/SwapRouter.sol
```diff
@@ -14,40 +17,54 @@ import "./interfaces/IVBNB.sol";
...
+    event SweepToken(address token, address to, uint256 sweepAmount);
+
     // *********************
     // **** CONSTRUCTOR ****
     // *********************
 
     /// @notice Constructor for the implementation contract. Sets immutable variables.
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor(address WBNB_, address factory_) RouterHelper(WBNB_, factory_) {
+    constructor(address WBNB_, address factory_, address _comptrollerAddress) RouterHelper(WBNB_, factory_) {
```
chechu, line 52: We have to update the deployment script to provide this address: `deploy/005-deploy-swaprouter.ts`

contracts/Swap/SwapRouter.sol
```diff
@@ -14,40 +17,54 @@ import "./interfaces/IVBNB.sol";
...
+    modifier ensureVTokenListed(address vTokenAddress) {
+        bool isListed = InterfaceComptroller(comptrollerAddress).markets(vTokenAddress);
+        if (isListed != true) {
+            revert VTokenNotListed(vTokenAddress);
+        }
+        _;
     }
 
+    /// @notice event emitted on sweep token success
+    event SweepToken(address token, address to, uint256 sweepAmount);
```
chechu, line 44: I would index the address fields

contracts/Swap/SwapRouter.sol
```diff
@@ -2,10 +2,13 @@ pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+
 import "./interfaces/IPancakeSwapV2Router.sol";
 import "./interfaces/IVtoken.sol";
 import "./RouterHelper.sol";
 import "./interfaces/IVBNB.sol";
+import "./interfaces/InterfaceComptroller.sol";
```
chechu, line 11: Can we reuse `ComptrollerInterface` defined here: `venus-protocol/contracts/Comptroller/ComptrollerInterface.sol`?

contracts/Swap/SwapRouter.sol
```diff
@@ -2,10 +2,13 @@ pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+
 import "./interfaces/IPancakeSwapV2Router.sol";
 import "./interfaces/IVtoken.sol";
```
chechu, line 8: VENUS-003 suggest to reuse `VTokenInterface` from `venus-protocol/contracts/Tokens/VTokens/VTokenInterfaces.sol`. What do you think about it?

contracts/Swap/SwapRouter.sol
defcon022: We cannot inherit interface from `ComptrollerInterface.sol` as the solidity version `0.5.16` is used and in SwapRouter it's `0.8.13`. We can take this part as future refactoring.

contracts/Swap/SwapRouter.sol
defcon022: Fixed.

deploy/005-deploy-swaprouter.ts
```diff
@@ -5,10 +5,12 @@ const ADDRESSES = {
   bsctestnet: {
     WBNBAddress: "0xae13d989daC2f0dEbFf460aC112a837C89BAa7cd",
     pancakeFactory: "0x182859893230dC89b114d6e2D547BFFE30474a21",
+    comptrollerAddress: "0x94d1820b2D1c7c7452A163983Dc888CEC546b77D",
```
chechu, line 8: The Comptroller is deployed in the same package. Would it be more flexible if we recover the Comptroller address dynamically, as we are doing [here](https://github.com/VenusProtocol/venus-protocol/blob/develop/deploy/004-support-markets.ts#L12)

defcon022: Let's do these changes in separate PR. As these fixes are related to the swap audits.

defcon022: Fixed.

contracts/Swap/SwapRouter.sol
```diff
@@ -2,10 +2,13 @@ pragma solidity 0.8.13;
 
 import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
```
chechu, line 4: We can remove this import now

contracts/Swap/SwapRouter.sol
```diff
@@ -14,22 +17,32 @@ import "./interfaces/IVBNB.sol";
  * @author 0xlucian
  */
 
-contract SwapRouter is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, RouterHelper, IPancakeSwapV2Router {
+contract SwapRouter is Ownable2StepUpgradeable, RouterHelper, IPancakeSwapV2Router {
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    address public comptrollerAddress;
```
chechu, line 23: In the future, if we want to make this SwapRouter contract compatible with IL, I think we'll need to replace this address with the PoolRegistry. But we can do it later. Even we could move this swap features to a new repo...

chechu (approved):

contracts/Swap/SwapRouter.sol
defcon022: Removed.

contracts/Swap/interfaces/IPancakeSwapV2Router.sol
```diff
@@ -1,75 +1,6 @@
 pragma solidity 0.8.13;
 
 interface IPancakeSwapV2Router {
-    function swapExactTokensForTokens(
```
chechu, line 4: We need to keep these "redundant" functions because our frontend will interact only with this SwapRouter contract, not directly with PCS

deploy/005-deploy-swaprouter.ts
```diff
@@ -20,13 +20,22 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const networkName = network.name === "bscmainnet" ? "bscmainnet" : "bsctestnet";
   const WBNBAddress = ADDRESSES[networkName].WBNBAddress;
   const pancakeFactoryAddress = ADDRESSES[networkName].pancakeFactory;
+  const comptroller = await deployments.get("Comptroller");
```
kkirka, line 23: Shouldn't this be `Unitroller`? AFAIU Comptroller is the implementation contract in core

tests/hardhat/Swap/swapTest.ts
```diff
@@ -440,6 +463,16 @@ describe("Swap Contract", () => {
...
     });
 
+    it("swap tokenA -> full debt of tokenB", async () => {
+      const deadline = await getValidDeadline();
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      await expect(
+        swapRouter
+          .connect(user)
+          .swapTokensForFullTokenDebtAndRepay(vToken.address, SWAP_AMOUNT, [tokenA.address, tokenB.address], deadline),
+      );
```
kkirka, line 473: Shouldn't we expect some result in this case? If we don't expect a result, this seems to be equivalent to `await swapRouter.connect(user).swapTokensForFullTokenDebtAndRepay(...)` 🤔

I guess this is the reason eslint reports a floating promise here

deploy/005-deploy-swaprouter.ts
aditya-routv: As far as I can tell, comptroller is utilised in every script and is deployed in '001-comptroller.ts' without unitroller. :thinking:

tests/hardhat/Swap/swapTest.ts
defcon022: In other tests also we are just verifying that they are not throwing any error, as we are using smock's fake contracts so we are unable to emit events.

deploy/005-deploy-swaprouter.ts
chechu: @0xlucian ?

chechu: In this case my proposal would be to do something like this:

```
import { Contracts as Mainnet } from "../networks/mainnet.json";
//TESTNET DEPLOYED CONTRACTS
import { Contracts as Testnet } from "../networks/testnet.json";
//MAINNET DEPLOYED CONTRACTS

[...]
interface AddressConfig {
  [key: string]: Address;
}

const addresses: AddressConfig = {
  bsctestnet: {
    unitroller: Testnet.Unitroller,
  },
  bscmainnet: {
    unitroller: Mainnet.Unitroller,
  },
};

[...]

const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
[...]
  const unitroller = await ethers.getContractAt("Unitroller", addresses[networkName].unitroller);
[...]
```

Similar to how we are dealing with these scenarios in `venus-protocol/deploy/005-update-governance.ts`

defcon022: @chechu chnages are done. can varify at 8ef98b63dc30e0f9e0f79a989d44950ea4a41ad4

UrAvgDeveloper (approved):

deploy/005-deploy-swaprouter.ts
```diff
@@ -1,14 +1,25 @@
...
+    unitroller: Testnet.Unitroller,
   },
   bscmainnet: {
-    WBNBAddress: "0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c",
-    pancakeFactory: "0xcA143Ce32Fe78f1f7019d7d551a6402fC5350c73",
+    WBNBAddress: Mainnet.WBNB,
+    pancakeFactory: Mainnet.pancakeFactory,
+    unitroller: Mainnet.Unitroller,
   },
 };
```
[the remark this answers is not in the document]
defcon022: @chechu, Fixed and pushed.

PR #233, opened 2023-03-20
[VEN-1238] Deny Claim/Deposit for Pending Withdrawal Request

narayanprusty:
## Description

Right now if a user has a pending withdrawal before upgrade then the deposit/claim revert due to underflow in calculations. So to get around this the user needs to execute the pending withdrawals first.

The fix is to enhance the experience by blocking claim/deposit call when there is a before upgrade withdrawal request pending and reverting with a proper error message.

## Checklist

### DIFF
--- a/contracts/XVSVault/XVSVault.sol
+++ b/contracts/XVSVault/XVSVault.sol
@@ -174,6 +174,11 @@ contract XVSVault is XVSVaultStorage, ECDSA {
         PoolInfo storage pool = poolInfos[_rewardToken][_pid];
         UserInfo storage user = userInfos[_rewardToken][_pid][msg.sender];
         _updatePool(_rewardToken, _pid);
+        require(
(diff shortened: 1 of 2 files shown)


chechu (approved):

aditya-routv (approved):
LGTM !

narayanprusty:
> Can we try to optimize the revert messages? They are bigger than 32 bytes currently and this will result in a bigger gas consumption. Lets either reduce the revert message to 32chars or just revert with custom error? wdyt?

@0xlucian In version 0.5.16 custom errors aren't supported. Also once we upgrade the version we will need to add custom errors to all reverts in the contract and not just the ones in this PR so that we maintain consistency. Therefore I think smaller length error messages for now will be the quickest way to go ahead. WDYT?

defcon022 (approved):

kkirka (approved):

PR #236, opened 2023-03-24
[VEN-1258] [VEN-1262]: vault pausable upgrades

GitGuru7:
## Description

~~This PR is currently in progress, it depends on new governance-contracts repo.~~
**XVSVault**
Make the XVSVault “pausable”.
pause and resume functions
integrate it with ACM, so, we’ll be able to execute them via Critical VIP’s
> deposit
> claim
> executeWithdrawal
> requestWithdrawal
> updatePool
> delegate

**VAIVault**
pause and resume functions
integrate it with ACM, so, we’ll be able to execute them via Critical VIP’s

> deposit
> withdraw
> claim
> updatePendingRewards

**VRTVault**
Change the code of the following functions to use the ACM:
> pause
> resume
> withdrawBep20

Stop rewarding with VRT from a specific block number: implemented a new state variable `lastAccruingBlock` and following function logic is extended:
> computeAccruedInterest

Added function to set `lastAccruingBlock` which is guarded by ACM
Added a new event `LastAccruingBlockChanged`

Resolves #

## Checklist


contracts/Vault/VAIVaultStorageV8.sol
```diff
@@ -0,0 +1,59 @@
...
+        uint256 rewardDebt;
+    }
+
+    // Info of each user that stakes tokens.
+    mapping(address => UserInfo) public userInfo;
+}
+
+contract VAIVaultStorageV1 is VAIVaultStorage {
+    /// @notice pause indicator for Vault
+    bool public vaultPaused;
```
chechu, line 58: Shouldn't we have to add this new state variable to`venus-protocol/contracts/Vault/VAIVaultStorage.sol` too?

tests/hardhat/Fork/VRTVaultUpgrade.ts
```diff
@@ -0,0 +1,172 @@
...
+  const vrtFactory = await ethers.getContractFactory("VRT");
+  vrt = await vrtFactory.deploy(deployer.address);
+
+  const vrtVaultFactory = await ethers.getContractFactory("VRTVault");
+  vrtVaultFresh = await vrtVaultFactory.deploy();
+  await vrtVaultFresh.initialize(vrt.address, bigNumber18);
+  await vrtVaultFresh._setAccessControl(ACM);
+}
+
+async function deployAndConfigureOldVault() {
```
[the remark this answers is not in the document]
aditya-routv: That's what we are doing here, connecting deployed vaultProxy address to factory, to maintain uniformity we have created two functions, `deployAndConfigureOldVault` and `deployAndConfigureNewVault`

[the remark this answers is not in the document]
aditya-routv: Yeah right, In `deployAndConfigureOldVault` function we are just forking block contain old implementation and making a instance of it by connecting it with factory and saving the storage states before the upgrade of new one to compare it with after upgrade.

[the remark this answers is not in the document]
aditya-routv: Yeah right, In `deployAndConfigureOldVault` function we are just forking block contain old implementation and making a instance of it by connecting it with factory and saving the storage states before the upgrade of new one to compare it with after upgrade.

contracts/Vault/VAIVault.sol
chechu, line 1: I would say the following functions can be `external` instead of `public`:
* deposit
* withdraw
* claim()
* claim(address account)
* updatePendingRewards
* getAdmin
* burnAdmin
* setNewAdmin
* _become
* setVenusInfo

Were these functions set `public` because of the tests?

contracts/VRTVault/VRTVault.sol
chechu, line 1: I would say the following functions should be `external` instead of `public`:
* pause
* resume

contracts/Vault/VAIVault.sol
aditya-routv: Fixed [here](https://github.com/VenusProtocol/venus-protocol/pull/236/commits/bd3db64bbe1940cf7386fe95d05c12479ce774a9)

defcon022 (approved):

tests/hardhat/Fork/swapTest.ts
chechu, line 1: Why is this file affected in this PR?

chechu (commented):
General comment: I would fix the solidity version in the `pragma` statement

contracts/Governance/IAccessControlManagerV5.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity ^0.5.16;
+
```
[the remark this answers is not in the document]
aditya-routv: While inheritance in interfaces is supported starting with Solidity version 6, I removed it at the time of the hotfix. Now adding back all functions from `IAccessControlV5` to `IAccessControlManagerV5`.

tests/hardhat/Fork/swapTest.ts
aditya-routv: I think it's because @0xlucian made some changes [here](https://github.com/VenusProtocol/venus-protocol/pull/236/commits/dc4903288abf55c382f00b1bcddab7e722e5877c) regarding syntactic sugar forking.

contracts/XVSVault/XVSVault.sol
```diff
@@ -4,18 +4,24 @@ pragma experimental ABIEncoderV2;
...
 }
 
-contract XVSVault is XVSVaultStorage, ECDSA {
+interface IXVSVaultProxy {
+    function _acceptImplementation() external returns (uint);
+
+    function admin() external returns (address);
+}
+
+contract XVSVault is XVSVaultStorageV2, ECDSA, AccessControlledV5 {
```
kkirka, line 24: If we use multiple inheritance, we need a `__gap` in `XVSVaultStorage`, otherwise the further upgrades will overwrite the storage of `AccessControlledV5`.

As a side note, with gaps using versioned storage becomes kinda tricky, so maybe we can just add new entries (`vaultPaused` and `__gap`) to `XVSVaultStorage`? Not sure here.

contracts/Governance/AccessControlledV5.sol
```diff
@@ -0,0 +1,76 @@
...
+import "./IAccessControlManagerV5.sol";
+
+/**
+ * @title Venus Access Control Contract.
+ * @dev This contract is helper between access control manager and actual contract
+ * This contract further inherited by other contract to integrate access controlled mechanism
+ * It provides initialise methods and verifying access methods
+ */
+
+contract AccessControlledV5 is InitializableV5, Ownable2StepUpgradeableV5 {
```
kkirka, line 15: I'm thinking whether we really need `Ownable2StepUpgradeableV5` here – AFAIU the admins of the vaults are set already, so this new `owner` field would live in parallel with the existing one. That's why, AFAIU, you needed a new `_setAccessControl` method in XVSVault rather than used `setAccessControlManager` defined in this contracts – it's because the owners of AccessControlledV5 and vaults are different. I understand the intent here, and I think making use of common base contracts is a good thing in general, but IMO having two distinct owners of the same contract will cause confusion.

contracts/Governance/AccessControlledV5.sol
```diff
@@ -0,0 +1,76 @@
...
+    event NewAccessControlManager(address oldAccessControlManager, address newAccessControlManager);
+
+    function __AccessControlled_init(address accessControlManager_) internal onlyInitializing {
+        __Ownable2Step_init();
+        __AccessControlled_init_unchained(accessControlManager_);
+    }
+
+    function __AccessControlled_init_unchained(address accessControlManager_) internal onlyInitializing {
+        _setAccessControlManager(accessControlManager_);
+    }
```
kkirka, line 36: We don't initialize this base contract in the vault contracts, so maybe we could simplify this V5 mixin by removing the initialization altogether? If we strip the initializers, we could remove some libraries we depend on as well.

kkirka (changes requested):
I think we need a team-wide discussion about how we use multiple inheritance and oz upgradeability/ownership in conjunction with the existing proxies, and maybe come up with some guidelines for the new code. These are quite complex/dangerous thingies that were not used in the original core code until now, so @VenusProtocol/contract I hope we could align on these here.

contracts/Governance/AccessControlledV5.sol
```diff
@@ -26,31 +24,6 @@ contract AccessControlledV5 is InitializableV5, Ownable2StepUpgradeableV5 {
...
-    function setAccessControlManager(address accessControlManager_) external onlyOwner {
-        _setAccessControlManager(accessControlManager_);
-    }
-
-    /**
-     * @notice Returns the address of the access control manager contract
-     */
-    function accessControlManager() external view returns (IAccessControlManagerV5) {
-        return _accessControlManager;
-    }
```
kkirka: Sorry for the misleading Slack post, maybe we could keep this external view for the transparency purposes, it won't harm IMO

kkirka (approved):
Left a small comment about the view, other than that LGTM

chechu (approved):

contracts/VRTVault/VRTVault.sol
```diff
@@ -63,13 +72,15 @@ contract VRTVault is VRTVaultStorage {
         _;
     }
 
-    function pause() public onlyAdmin {
+    function pause() external {
```
coreyar, line 75: External and public functions should have natspec

coreyar: These functions should also come after the modifiers

GitGuru7: Fixed [here](https://github.com/VenusProtocol/venus-protocol/pull/236/commits/0b23a7600272fd7b1b932845d5c5b914a567d909)

PR #254, opened 2023-04-25
[VEN-1380] Optimised swap router to reduce the size under 24 KB

defcon022:
## Description

Resolves #

## Checklist

### DIFF
--- a/contracts/Swap/SwapRouter.sol
+++ b/contracts/Swap/SwapRouter.sol
@@ -667,10 +667,10 @@ contract SwapRouter is Ownable2Step, RouterHelper, IPancakeSwapV2Router {
     }
 
     /**
-     * @notice supply token to a Venus market
-     * @param path the addresses of the underlying token
+     * @notice Supply token to a Venus market
+     * @param path The addresses of the underlying token
      * @param vTokenAddress The address of the vToken contract for supplying assets.
-     * @param swapAmount the amount of tokens supply to Venus Market.
+     * @param swapAmount The amount of tokens supply to Venus Market.
      */
     function _supply(address path, address vTokenAddress, uint256 swapAmount) internal {
         TransferHelper.safeApprove(path, vTokenAddress, 0);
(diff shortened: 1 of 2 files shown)


contracts/Swap/SwapRouter.sol
```diff
@@ -717,4 +695,31 @@ contract SwapRouter is Ownable2Step, RouterHelper, IPancakeSwapV2Router {
             revert RepayError(msg.sender, vTokenAddress, response);
         }
     }
+
+    /**
+     * @notice check for min amount out.
+     * @param path The addresses of the underlying token
+     * @param balanceBefore Balance before the swap.
+     * @param amountOutMin Min amount out threshold.
```
chechu, line 703: I miss the `to` param doc. The `@notice` fragment could be: check if the balance of `to` minus the balanceBefore is greater than the amountOutMin

contracts/Swap/SwapRouter.sol
```diff
@@ -717,4 +695,31 @@ contract SwapRouter is Ownable2Step, RouterHelper, IPancakeSwapV2Router {
...
+        uint256 amountOutMin,
+        address to
+    ) internal view returns (uint256 swapAmount) {
+        uint256 balanceAfter = IERC20(path[path.length - 1]).balanceOf(to);
+        swapAmount = balanceAfter - balanceBefore;
+        require(swapAmount >= amountOutMin, "SwapRouter: SwapAmount is less than amountOutMin");
+    }
+
+    /**
+     * @notice Get swapped amount
```
chechu, line 717: We should try to write better docs. Here something like "returns the difference between the balance of `this` and the balanceBefore" could be better

contracts/Swap/SwapRouter.sol
defcon022: Fixed, 6ef9faf8fc19a066c7e4a7dac440a9bd25f2a1cc

contracts/Swap/SwapRouter.sol
defcon022: Fixed, 6ef9faf8fc19a066c7e4a7dac440a9bd25f2a1cc

chechu (approved):

PR #259, opened 2023-05-04
[VEN-1148]: send reserves to protocol share reserve

GitGuru7:
## Description

Liquidator : 

- Redeem the VTokens assigned to the protocol, ensuring that the redeem won’t revert the liquidation transaction (i.e. because there are not enough liquidity to complete the redeem, or because the redeem action is disabled)
- send the underlying tokens to the `ProtocolShareReserve` contract, identifying the income origin (`liquidation`)
- have a function, externally invocable by anyone, that would transfer the pending accumulated amounts. This will allow us to “force” manually the transfers, if needed

If redeem fails : 

- The Liquidator contract should store in a storage variable the address of the VToken not redeemed
- At the end of each liquidation, the Liquidator contract will try to redeem the pending VTokens, stored in the previous list (set)
- In the set of pending VTokens to redeem we only need to store the VToken addresses. The balances can be calculated each time. Each time a pending VToken is redeemed, it’s removed from the pending list

Resolves VEN-1148

## Checklist


contracts/Liquidator/Liquidator.sol
```diff
@@ -305,6 +311,43 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Acce
...
+
+    function _reduceReservesInternal() internal {
+        for (uint256 index = pendingRedeem.length - 1; index > 0; index--) {
+            address vToken = pendingRedeem[index];
+            uint256 vTokenBalance = IVToken(vToken).balanceOf(address(this));
+            uint256 redeemResponse = IVBep20(vToken).redeem(vTokenBalance);
+            address underlying = IVBep20(vToken).underlying();
+
+            if (redeemResponse == 0) {
+                uint256 underlyingBalance = IVToken(underlying).balanceOf((address(this)));
```
chechu, line 338: Just to avoid any confusion...
```suggestion
                uint256 underlyingBalance = IERC20Upgradeable(underlying).balanceOf((address(this)));
```

contracts/Liquidator/Liquidator.sol
```diff
@@ -305,6 +311,43 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Acce
...
+                if (!IVToken(underlying).transfer(protocolShareReserve, underlyingBalance)) {
```
chechu, line 339: Two points here:
* I would use `IERC20Upgradeable` instead `IVToken`, to avoid confusions
* We should take into account the transfer of BNB

contracts/Liquidator/Liquidator.sol
```diff
@@ -40,7 +43,11 @@ interface IVAIController {
     function getVAIRepayAmount(address borrower) external view returns (uint256);
 }
 
-contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, AccessControlledV8 {
+interface IProtocolShareReserve {
+    function updateAssetsState(address comptroller, address asset, uint256 kind) external;
```
chechu, line 47: I would define the enum with the considered kind of origins, and I would use it in this file. It will be duplicated in other files, but at least it won't be a magic number. Alternatively, we can use constants

contracts/Liquidator/Liquidator.sol
```diff
@@ -22,6 +23,8 @@ interface IVBep20 is IVToken {
         uint256 repayAmount,
         IVToken vTokenCollateral
     ) external returns (uint256);
+
+    function redeem(uint256 redeemTokens) external returns (uint256);
```
chechu, line 27: `redeem` is still available in `vBNB`. I would move it to the `IVToken` interface

contracts/Liquidator/Liquidator.sol
```diff
@@ -305,6 +311,43 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Acce
...
+     */
+    function reduceReserves() external {
+        _reduceReservesInternal();
+    }
+
+    function _reduceReservesInternal() internal {
+        for (uint256 index = pendingRedeem.length - 1; index > 0; index--) {
+            address vToken = pendingRedeem[index];
+            uint256 vTokenBalance = IVToken(vToken).balanceOf(address(this));
+            uint256 redeemResponse = IVBep20(vToken).redeem(vTokenBalance);
```
chechu, line 334: I would use `IVToken` instead of `IVBep20`, to avoid confusions

contracts/Liquidator/Liquidator.sol
```diff
@@ -305,6 +311,43 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Acce
...
+            address underlying = IVBep20(vToken).underlying();
```
chechu, line 335: If the vToken is vBNB, `underlying()` does not exist

contracts/Liquidator/Liquidator.sol
```diff
@@ -334,8 +377,17 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Acce
...
+        if (IVBep20(address(vTokenCollateral)).redeem(ours) != 0) {
+            pendingRedeem.push(address(vTokenCollateral));
+        } else {
+            address underlying = IVBep20(address(vTokenCollateral)).underlying();
+            uint256 underlyingBalance = IVToken(underlying).balanceOf((address(this)));
+            if (!IVToken(underlying).transfer(protocolShareReserve, underlyingBalance)) {
+                revert UnderlyingTransferFailed(address(vTokenCollateral), underlying);
+            }
+            IProtocolShareReserve(protocolShareReserve).updateAssetsState(address(comptroller), underlying, 1);
+            emit ReservesReduced(msg.sender, address(vTokenCollateral), ours);
```
chechu, line 390: I would use functions to reuse this sequence of statements in the `_reduceReservesInternal()`

chechu (commented):
We should take into account the collateral can be vBNB

chechu (commented):
For BNB liquidations, I would wrap it to WBNB before transferring them to the `ProtocolSharedReserves`, and I would transfer the WBNB tokens. This way, the process will be more simple downstream

contracts/Liquidator/Liquidator.sol
```diff
@@ -381,17 +398,43 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
+            address(vBnb),
+            IncomeType.LIQUIDATION
+        );
+        emit ReservesReduced(msg.sender, address(wBNB), wBnbBalance);
+    }
+
+    function _reduceVTokenReserves(address vToken) private {
+        address underlying = IVBep20(vToken).underlying();
+        uint256 underlyingBalance = IERC20Upgradeable(underlying).balanceOf((address(this)));
+        if (!IERC20Upgradeable(underlying).transfer(protocolShareReserve, underlyingBalance)) {
```
chechu, line 431: Would it be better if we use `SafeERC20` and `safeTranfer`? (as we are doing in the SwapRouter case [here](https://github.com/VenusProtocol/venus-protocol/blob/develop/contracts/Swap/SwapRouter.sol#L20))

contracts/Liquidator/Liquidator.sol
```diff
@@ -381,17 +398,43 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
+            revert UnderlyingTransferFailed(vToken, underlying);
+        }
+        IProtocolShareReserve(protocolShareReserve).updateAssetsState(
+            address(comptroller),
+            underlying,
```
chechu, line 436: `vToken`?

contracts/Liquidator/Liquidator.sol
```diff
@@ -147,18 +162,24 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Acce
...
     using SafeERC20Upgradeable for IERC20Upgradeable;
 
     /// @notice Constructor for the implementation contract. Sets immutable variables.
     /// @param comptroller_ The address of the Comptroller contract
     /// @param vBnb_ The address of the VBNB
     /// @param treasury_ The address of Venus treasury
+    /// @param wBNB_ The address of wBNB
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor(address comptroller_, address payable vBnb_, address treasury_) {
+    constructor(address comptroller_, address payable vBnb_, address treasury_, address wBNB_) {
```
chechu, line 176: Shouldn't we include the `wBNB_` in the `initialize` function instead to allow us to set it up during the re-initialization?

contracts/Liquidator/Liquidator.sol
GitGuru7: According to the present configuration in [IL](https://github.com/VenusProtocol/isolated-pools/blob/develop/contracts/VToken.sol#L1233), assets are only identified as underlying.
Should we also include vToken?

contracts/Liquidator/Liquidator.sol
GitGuru7: Instead of setting it at re-initialization, why not do it during deployment and keep it immutable like the vBNB variable Wdys?

contracts/Liquidator/Liquidator.sol
chechu: You are right, my fault

contracts/Liquidator/Liquidator.sol
```diff
@@ -90,10 +111,16 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
 
     /// @notice Emitted when force liquidation is resumed
-    event ForceVaiLiquidationResumed(address sender);
+    event ForceVAILiquidationResumed(address indexed sender);
+
+    /// @notice Emitted when new address of protocol share reserve is set
+    event NewProtocolShareReserve(address indexed oldProtocolShareReserve, address indexed newProtocolShareReserves);
+
+    /// @notice Emitted when reserves are reduced from liquidator contract to protocol share reserves
+    event ReservesReduced(address indexed sender, address indexed token, uint256 reducedAmount);
```
chechu, line 123: I would use a different name for this event. I assume you chose this name to make it equal to the event `ReservesReduced` emitted in the VToken contracts, but I would be explicit with the context where this is happening and I would name it something like: `ProtocolLiquidationIncentiveTransferred(address indexed liquidator, address indexed token, uint256 amount)`

contracts/Liquidator/Liquidator.sol
```diff
@@ -289,6 +331,39 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
+        _reduceReservesInternal();
+    }
+
+    function _reduceReservesInternal() internal {
+        uint256 pendingReedemLength_ = pendingRedeem.length;
+        if (pendingReedemLength_ == 0) return;
+        for (uint256 index = pendingReedemLength_ - 1; index >= 0; index--) {
+            address vToken = pendingRedeem[index];
+            uint256 vTokenBalance_ = IVToken(vToken).balanceOf(address(this));
+            if (IVToken(vToken).redeem(vTokenBalance_) != 0) {
```
chechu, line 355: If `redeem` returned value is not 0, then there was an error. I would say, here we have to check `==`, instead of `!=`

contracts/Liquidator/Liquidator.sol
```diff
@@ -318,12 +392,45 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
+            address(comptroller),
+            wBNB,
+            IncomeType.LIQUIDATION
+        );
+        emit ReservesReduced(msg.sender, address(wBNB), bnbBalance);
+    }
+
+    function _reduceVTokenReserves(address vToken) private {
+        address underlying = IVBep20(vToken).underlying();
+        uint256 underlyingBalance = IERC20Upgradeable(underlying).balanceOf((address(this)));
```
chechu, line 424:
```suggestion
        uint256 underlyingBalance = IERC20Upgradeable(underlying).balanceOf(address(this));
```

contracts/Liquidator/Liquidator.sol
```diff
@@ -318,12 +392,45 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
+        IERC20Upgradeable(underlying).safeTransfer(protocolShareReserve, underlyingBalance);
```
chechu, line 425: I would add support to deflationary/rebase tokens, adding a function `doTransferOut` that will return the really transferred amount. So, we can include the real tokens transferred in the event

contracts/Liquidator/Liquidator.sol
```diff
@@ -7,13 +7,20 @@ import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 import "@venusprotocol/governance-contracts/contracts/Governance/AccessControlledV8.sol";
 import "./LiquidatorStorage.sol";
 
+enum IncomeType {
+    SPREAD,
+    LIQUIDATION
+}
```
kkirka, line 13: I feel like this enum belongs somwhere else, not sure where though

contracts/Liquidator/Liquidator.sol
kkirka:
```
uint256 internal constant NO_ERROR = 0;
...
if (IVToken(vToken).redeem(vTokenBalance_) == NO_ERROR) ...
```

contracts/Liquidator/Liquidator.sol
```diff
@@ -318,12 +392,45 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
         }
         return (ours, theirs);
     }
 
+    function _reduceBnbReserves() private {
+        uint256 bnbBalance = address(this).balance;
+        IWBNB(wBNB).deposit{ value: bnbBalance }();
+        if (!IWBNB(wBNB).transfer(protocolShareReserve, bnbBalance)) {
+            revert UnderlyingTransferFailed(address(vBnb), wBNB);
+        }
```
kkirka, line 413: Maybe just
```suggestion
        IERC20Upgradeable(wBNB).safeTransfer(protocolShareReserve, bnbBalance);
```

contracts/Liquidator/Liquidator.sol
GitGuru7: Correct I missed equality check.

contracts/Liquidator/Liquidator.sol
GitGuru7: To avoid any confusion I think we can use transfer for `wBNB` transfers.....also `wBNB` [contract](https://bscscan.com/address/0xbb4CdB9CBd36B01bD1cBaEBF2De08d9173bc095c#code) does not have `safeTransfer` WDYT?

contracts/Liquidator/Liquidator.sol
chechu: We can keep it here "today" and import it from the new package in the near future

contracts/Liquidator/Liquidator.sol
chechu: We can keep it as it is. We are only using the amount for the event, and the `updateAssetsState` function is emitting the real transferred amount. So, we can save some gas not doing the extra calculation

contracts/Liquidator/Liquidator.sol
```diff
@@ -289,6 +331,39 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
...
+        for (uint256 index = pendingReedemLength_ - 1; index >= 0; index--) {
+            address vToken = pendingRedeem[index];
+            uint256 vTokenBalance_ = IVToken(vToken).balanceOf(address(this));
+            if (IVToken(vToken).redeem(vTokenBalance_) != 0) {
+                if (vToken == address(vBnb)) {
+                    _reduceBnbReserves();
+                } else {
+                    _reduceVTokenReserves(vToken);
+                }
+                pendingRedeem[index] = pendingRedeem[pendingReedemLength_ - 1];
```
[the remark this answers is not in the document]
GitGuru7: Fixed [here](https://github.com/VenusProtocol/venus-protocol/pull/259/commits/597a2ec13591bbbb92f06343fee721180e90db2c)

contracts/Liquidator/Liquidator.sol
GitGuru7: Fixed [here](https://github.com/VenusProtocol/venus-protocol/pull/259/commits/597a2ec13591bbbb92f06343fee721180e90db2c)

contracts/Liquidator/Liquidator.sol
GitGuru7: We must also take into account the possibility that the liquidation transaction will be reverted if the market action `redeem` is paused.

contracts/Liquidator/Liquidator.sol
GitGuru7: Fixed [here](https://github.com/VenusProtocol/venus-protocol/pull/259/commits/597a2ec13591bbbb92f06343fee721180e90db2c)

contracts/Liquidator/Liquidator.sol
```diff
@@ -352,16 +352,16 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
     function _reduceReservesInternal() internal {
         uint256 pendingReedemLength_ = pendingRedeem.length;
         if (pendingReedemLength_ == 0) return;
-        for (uint256 index = pendingReedemLength_ - 1; index >= 0; index--) {
-            address vToken = pendingRedeem[index];
+        for (int256 index = int256(pendingReedemLength_) - 1; index >= 0; index--) {
```
chechu, line 355: Let's say `pendingReedemLength_` is high enough so we consume 100% of the gas in the loop. Would it make sense to cap this value in this for-loop, so we would process always `min(10, pendingReedemLength_)`, for example. This way we would ensure we reduce some reserves in each interaction

contracts/Liquidator/Liquidator.sol
```diff
@@ -351,8 +351,9 @@ contract Liquidator is Ownable2StepUpgradeable, ReentrancyGuardUpgradeable, Liqu
 
     function _reduceReservesInternal() internal {
         uint256 pendingReedemLength_ = pendingRedeem.length;
+        uint256 chunk = pendingReedemLength_ >= 10 ? 10 : pendingReedemLength_;
```
chechu, line 354: 1. I would extract this number to a constant at least
2. I said 10 as an example. It would be great if we can estimate the gas spent at each iteration, to confirm 10 is good enough

chechu (approved):

tests/hardhat/Fork/reduceResevesTest.ts
```diff
@@ -0,0 +1,250 @@
...
+          .liquidateBorrow(borrowedToken1, borrower1, repayAmount1, collateralToken1, { value: repayAmount1 }),
+      ).to.be.emit(liquidator, "LiquidateBorrowedTokens");
+
+      await expect(
+        liquidator
+          .connect(liquidatorSigner2)
+          .liquidateBorrow(borrowedToken2, borrower2, repayAmount2, collateralToken2, { value: repayAmount2 }),
+      ).to.be.emit(liquidator, "LiquidateBorrowedTokens");
+
+      await comptroller._setActionsPaused([collateralToken1, collateralToken2], [1, 1], false);
```
chechu, line 238:
```suggestion
      await comptroller._setActionsPaused([collateralToken1, collateralToken2], [1], false);
```

contracts/Liquidator/Liquidator.sol
kkirka: But WBNB is ERC-20, so it should be possible to use SafeERC20 wrappers. I don't see any risk with `transfer`, but it would trigger automated analysis tools, so I'd use oz's wrapper to unify the approach.

PR #274, opened 2023-05-31
test: fix dangling promises in tests

kkirka:
Some of our tests have missing awaits. It has been flagged by linter but as a warning, not as an error. However, dangling promises may cause test suite to succeed when it shouldn't, e.g. missing await before expecting an event would make the test succeed even if the event is not emitted. This PR adds the missing awaits where appropriate.


GitGuru7 (approved):

kkirka:
UPD: Turned out some of the tests were failing, but we didn't know about that. Pushed some commits to fix the failing tests.

chechu (approved):

GitGuru7 (approved):

PR #289, opened 2023-06-16
[VEN-1474]: Harvest BNB Income

web3rover:
## Description

Resolves VEN-1474

## Checklist

### DIFF
--- a/tests/hardhat/Admin/VBNBAdmin.ts
+++ b/tests/hardhat/Admin/VBNBAdmin.ts
@@ -6,7 +6,7 @@ import { BigNumber, Signer } from "ethers";
 import {
   ComptrollerHarness,
   ComptrollerHarness__factory,
-  IAccessControlManager,
+  IAccessControlManagerV8,
   IProtocolShareReserve,
   InterestRateModelHarness,
   MockVBNB,
@@ -55,7 +55,7 @@ const setupMarketFixture = async (): Promise<SetupMarketFixture> => {
   const WBNBFactory = await smock.mock<WBNB__factory>("WBNB");
   const WBNB = await WBNBFactory.deploy();
 
-  const accessControl = await smock.fake<IAccessControlManager>("IAccessControlManager");
+  const accessControl = await smock.fake<IAccessControlManagerV8>("IAccessControlManagerV8");
   accessControl.isAllowedToCall.returns(true);
 
   const VBNBAdminFactory = await ethers.getContractFactory("VBNBAdmin");


contracts/Admin/VBNBAdminStorage.sol
```diff
@@ -0,0 +1,40 @@
...
+    IWBNB public WBNB;
+
+    /// @notice address of protocol share reserve contract
+    IProtocolShareReserve public protocolShareReserve;
+
+    /// @notice address of comptroller contract
+    address public comptroller;
+
+    /// @notice gap to prevent collision in inheritence
+    uint256[49] private __gap;
```
GitGuru7, line 39:
```suggestion
    /// @dev gap to prevent collision in inheritence
    uint256[49] private __gap;
```

contracts/Admin/VBNBAdminStorage.sol
```diff
@@ -0,0 +1,40 @@
...
+
+contract VBNBAdminStorage {
+    /// @notice address of vBNB contract
+    VTokenInterface public vBNB;
+
+    /// @notice address of WBNB contract
+    IWBNB public WBNB;
+
+    /// @notice address of protocol share reserve contract
+    IProtocolShareReserve public protocolShareReserve;
```
GitGuru7, line 33: I think we would require a setter for it.

contracts/Admin/VBNBAdmin.sol
```diff
@@ -0,0 +1,96 @@
...
+        IWBNB _WBNB,
+        address _comptroller,
+        address accessControlManager
+    ) external initializer {
+        vBNB = _vBNB;
+        protocolShareReserve = _protocolShareReserve;
+        WBNB = _WBNB;
+        comptroller = _comptroller;
+
+        __Ownable2Step_init();
```
chechu, line 32: Redundant: it's already invoked in `__AccessControlled_init()`

contracts/Admin/VBNBAdmin.sol
```diff
@@ -0,0 +1,96 @@
...
+        VTokenInterface _vBNB,
+        IProtocolShareReserve _protocolShareReserve,
+        IWBNB _WBNB,
+        address _comptroller,
+        address accessControlManager
+    ) external initializer {
+        vBNB = _vBNB;
+        protocolShareReserve = _protocolShareReserve;
+        WBNB = _WBNB;
+        comptroller = _comptroller;
```
chechu, line 30: Would it make sense to assume that the comptroller will be `vBNB.comptroller()`? This way we can save a storage slot

chechu (approved):

contracts/Admin/VBNBAdmin.sol
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: BSD-3-Clause
+
+pragma solidity 0.8.13;
+
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
Debugger022, line 6:
```suggestion
```

Debugger022 (approved):

tests/hardhat/Admin/VBNBAdmin.ts
```diff
@@ -0,0 +1,124 @@
...
+  const protocolShareReserve = await smock.fake<IProtocolShareReserve>("IProtocolShareReserve");
+
+  const WBNBFactory = await smock.mock<WBNB__factory>("WBNB");
+  const WBNB = await WBNBFactory.deploy();
+
+  const accessControl = await smock.fake<IAccessControlManager>("IAccessControlManager");
+  accessControl.isAllowedToCall.returns(true);
+
+  const VBNBAdminFactory = await ethers.getContractFactory("VBNBAdmin");
+  const VBNBAdmin: VBNBAdmin = await upgrades.deployProxy(VBNBAdminFactory, [
```
andy-b-b, line 62: It is suggested to test this against the live vBNB and network.
I guess this does not work because:
1. `transfer` has 2300 gas limit. VBNBAdmin might not be able to receive BNB
2. vBNB is not upgradable.

https://github.com/VenusProtocol/venus-protocol/blob/develop/contracts/Tokens/VTokens/VBNB.sol.archive#L147

andy-b-b: I did came across this issue before(`TransparentUpgradeableProxy`, that's the default).

Here is a discussion(from other users).
https://forum.openzeppelin.com/t/transparentupgradeableproxy-is-there-any-risk-in-omitting-admin-check-from-receive-function/27315

contracts/Admin/VBNBAdmin.sol
```diff
@@ -0,0 +1,121 @@
...
+    function _wrapBNB() internal {
+        uint256 bnbBalance = address(this).balance;
+        WBNB.deposit{ value: bnbBalance }();
+    }
+
+    /**
+     * @notice Invoked when BNB is sent to this contract
+     */
+    receive() external payable {
+        require(msg.sender == address(vBNB), "only vBNB can send BNB to this contract");
```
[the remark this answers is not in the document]
andy-b-b: Got it. Thanks

contracts/Admin/VBNBAdmin.sol
```diff
@@ -0,0 +1,117 @@
...
+        require(address(protocolShareReserve_) != address(0), "PSR address invalid");
+        emit ProtocolShareReserveUpdated(protocolShareReserve, protocolShareReserve_);
+        protocolShareReserve = protocolShareReserve_;
+    }
+
+    /**
+     * @notice Reduce reserves of vBNB
+     * @param reduceAmount amount of reserves to reduce
+     */
+    function reduceReserves(uint reduceAmount) external nonReentrant {
```
andy-b-b, line 69: Just being curious.

Only authorised users can `VToken::_reduceReserves()`. However, anyone can do `VBNBAdmin::reduceReserves()`. Not sure we have a special consideration?

chechu: The goal is to make permisionless the "automatic income allocation". `VBNBAdmin::reduceReserves()` is going to transfer the reduced reserves to a pre-fixed address, set by Governance via VIP, so it should be safe and it's not needed to limit it to authorized accounts. VBNB is not upgradable, so the behavior of its `VToken::_reduceReserves()` cannot be changed. We think we are covering the requirements (permissionless) with the proposed solution.

deploy/007-deploy-VBNBAdmin.ts
```diff
@@ -0,0 +1,67 @@
...
+    from: deployer,
+    args: [vBNB, WBNB],
+    log: true,
+    autoMine: true,
+    proxy: {
+      owner: ADDRESSES[networkName].Timelock,
+      proxyContract: "OpenZeppelinTransparentProxy",
+      execute: {
+        methodName: "initialize",
+        args: [acmAddresses[networkName], treasuryAddresses[networkName]],
```
chechu, line 49: This should be:
* First: protocol share reserve
* Second: ACM

GitGuru7 (approved):

hardhat.config.ts
```diff
@@ -103,10 +103,13 @@ const config: HardhatUserConfig = {
       timeout: 12000000,
     },
     bscmainnet: {
-      url: "http://127.0.0.1:1248",
+      url: "https://bsc-dataseed.binance.org",
```
chechu, line 106: I wouldn't do this change. We should use Frame for the deployment of contracts to mainnet

deployments/bscmainnet/DefaultProxyAdmin.json
chechu, line 1: The previous DefaultProxyAdmin is the the ProxyAdmin used in the PSM contract. We could rename the old file to `PSMProxyAdmin`, to keep it in the repo. Maybe in the future we can propose a VIP to change the admin of the PSM proxy, to use the new one, but in the meantime at least we don't lose the track of this

chechu (approved):

hardhat.config.ts
```diff
@@ -103,10 +103,13 @@ const config: HardhatUserConfig = {
...
     },
     bscmainnet: {
-      url: "http://127.0.0.1:1248",
+      url: "https://bsc-dataseed.binance.org/",
       chainId: 56,
       live: true,
       timeout: 1200000, // 20 minutes
+      accounts: {
+        mnemonic: process.env.MNEMONIC || "",
+      },
```
chechu, line 112: Don't push this change

chechu (approved):

chechu (approved):

PR #318, opened 2023-08-22
Release

chechu:


chechu:
Replaced by https://github.com/VenusProtocol/venus-protocol/pull/319. There are changes in the `develop` branch that could break the integrations with the currently deployed contracts (like changes in emitted events), so it's better to be more conservative and only update the files with the addresses, that it's what we initially need

PR #362, opened 2023-10-05
[VEN-2005]: BUSD Liquidator

chechu:
## Description

This PR adds a custom mechanism for BUSD liquidations. Only BUSD liquidator contract will be allowed to liquidate BUSD borrows. To reduce the potential impact of this change, we will not upgrade Comptroller or Liquidator contracts. Instead, BUSD liquidator contract will be allowed to pause and unpause liquidating BUSD debts. Between the transactions, BUSD liquidations will be paused, as far as Comptroller is concerned.

Once a liquidation is triggered in the BUSD liquidator contract, it would:

1. Unpause the liquidations in the Comptroller
1. Transfer BUSD from the sender and approve the Venus Liquidator contract
1. Perform a liquidation
1. Distribute the seized collateral, including the liquidation incentive, according to a custom distribution scheme, configured by Governance

### DIFF
--- a/contracts/Liquidator/BUSDLiquidator.sol
+++ b/contracts/Liquidator/BUSDLiquidator.sol
@@ -1,0 +1,189 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.13;
+
+import { SafeERC20Upgradeable, IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import { ReentrancyGuardUpgradeable } from "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+import { Ownable2StepUpgradeable } from "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+
+import { MANTISSA_ONE } from "./constants.sol";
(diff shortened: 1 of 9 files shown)


contracts/Liquidator/BUSDLiquidator.sol
```diff
@@ -0,0 +1,188 @@
...
+    /// @dev Transfers tokens to this contract and returns the actual transfer amount
+    /// @param token The token to transfer
+    /// @param from The account to transfer from
+    /// @param amount The amount to transfer
+    /// @return actualAmount The actual amount transferred
+    function _transferIn(
+        IERC20Upgradeable token,
+        address from,
+        uint256 amount
+    ) internal returns (uint256 actualAmount) {
```
chechu, line 147: I think we don't need to name the returned variable, because we are using `return`

chechu (commented):
LGTM

contracts/Liquidator/BUSDLiquidator.sol
kkirka: Fixed in [a59de0b](https://github.com/VenusProtocol/venus-protocol/pull/362/commits/a59de0bcaa024ded6f728972d316ee5c880f0f4f)

kkirka (approved):

chechu (commented):
LGTM

PR #366, opened 2023-10-06
[VEN-2066][VEN-2109]: add XVS vault deployment script

coreyar:
## Description

Adds deployment script for the XVS vault


deploy/007-deploy-vaults.ts
```diff
@@ -0,0 +1,67 @@
...
+import { DeployFunction } from 'hardhat-deploy/types';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import deployedContracts from '@venusprotocol/governance-contracts/deployments/deployments.json'
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { deployments, getNamedAccounts } = hre;
+  const { deploy } = deployments;
+  const { deployer } = await getNamedAccounts();
+
+  const xvsDeployment = await deploy('XVS', {
```
chechu, line 11: I would deploy XVS in a previous script. This way we can define the address of the XVS token to be considered on each network. For example, in Ethereum the XVS will be the bridged XVS deployed [here](https://github.com/VenusProtocol/isolated-pools/pull/291)

deploy/008-deploy-vaults.ts
```diff
@@ -0,0 +1,64 @@
...
+        "AccessControlManager",
+        deployedContracts[chainId][0].contracts.AccessControlManager.address,
+      )
+    : await deployments.get("AccessControlManager");
+
+  let txn = await xvsVault.setXvsStore(xvs.address, xvsStore.address);
+  await txn.wait(1);
+
+  txn = await xvsVault.setAccessControl(accessControlManager.address);
+  await txn.wait(1);
```
kkirka, line 52: These should be called on proxy after _become. Otherwise you're setting values (including _notEntered) in the implementation storage which isn't used when you call functions via proxy

deploy/008-deploy-vaults.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  // Become Implementation of XVSVaultProxy
+  await xvsVaultProxy._setPendingImplementation(xvsVaultAddress);
+  await xvsVault._become(xvsVaultProxyAddress);
+
+  let txn = await xvsVault.setXvsStore(xvs.address, xvsStore.address);
+  await txn.wait(1);
+
+  txn = await xvsVault.setAccessControl(accessControlManager.address);
+  await txn.wait(1);
```
GitGuru7, line 56:
```suggestion
  let txn = await xvsVaultProxy.setXvsStore(xvs.address, xvsStore.address);
  await txn.wait(1);

  txn = await xvsVaultProxy.setAccessControl(accessControlManager.address);
  await txn.wait(1);
```

deploy/007-deploy-xvs.ts
```diff
@@ -0,0 +1,19 @@
...
+    from: deployer,
+    args: [deployer],
+    log: true,
+    autoMine: true,
+  });
+};
+
+func.tags = ["xvs"];
+
+export default func;
```
GitGuru7, line 19: We should skip deployment if network is live

deploy/008-deploy-vaults.ts
```diff
@@ -0,0 +1,64 @@
...
+  const xvsVaultProxyAddress = xvsVaultProxyDeployment.address;
+
+  await deploy("XVSStore", {
+    from: deployer,
+    args: [],
+    log: true,
+    autoMine: true,
+  });
+
+  const xvs = await ethers.getContract("XVS");
```
GitGuru7, line 36: We should fetch XVS address from deployments file because remote XVS token contract would be different from the current one we have on bsc

deploy/007-deploy-xvs.ts
coreyar: Rather than skipping it should we modify it to deploy the correct XVS contract based on the network and skip on BSC since it is already deployed there?

deploy/008-deploy-vaults.ts
coreyar: If we have the contract deployed and exported under this name wouldn't hardhat pick the correct deployment/ address based on the network?

GitGuru7: I think we need to create instance by using address from token-bridge package

deploy/007-deploy-xvs.ts
coreyar: Makes sense I already added skipping it on a live network

deploy/008-deploy-vaults.ts
coreyar: I've added the token bridge package to the external deployments in hardhat. This should pull the XVS address correctly from the deployments based on the network

deploy/009-configure-vaults.ts
```diff
@@ -0,0 +1,67 @@
+import deployedContracts from "@venusprotocol/governance-contracts/deployments/deployments.json";
```
chechu, line 1: I think this json file is not available in the package `@venusprotocol/governance-contracts`. Moreover, I would update the dependency in the package.json, to use the version `1.4.0-dev.2`, that includes the ACM deployed to Ethereum

deploy/009-configure-vaults.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const xvsVaultProxy = await ethers.getContractAt("XVSVault", xvsVaultProxyDeployment.address);
+
+  let txn = await xvsVaultProxy.setXvsStore(xvs.address, xvsStoreDeployment.address);
+  await txn.wait();
+
+  txn = await xvsVaultProxy.setAccessControl(accessControlManager.address);
+  await txn.wait();
+
+  if (!hre.network.live) {
```
chechu, line 38: `bsctestnet`, in hardhat.config.ts, doesn't have `live: true`. Should we set it?

deploy/009-configure-vaults.ts
coreyar: I chose to resolve this by adding deployments to the hardhat config. The governance-contracts repo version has also been bumped

PR #382, opened 2023-11-14
Add vtreasury deployment

coreyar:
## Description

Adds deployed vTreasury contract to deployments

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -101,7 +101,7 @@ const config: HardhatUserConfig = {
       timeout: 12000000,
     },
     bscmainnet: {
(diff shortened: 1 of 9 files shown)


hardhat.config.ts
```diff
@@ -103,7 +103,7 @@ const config: HardhatUserConfig = {
       timeout: 12000000,
     },
     bscmainnet: {
-      url: "http://127.0.0.1:1248",
+      url: process.env.BSC_ARCHIVE_NODE_URL || "https://bsc-dataseed.binance.org/",
```
chechu, line 106: I would keep `http://127.0.0.1:1248`, to force us to use Frame in the deployments to mainnet

PR #404, opened 2023-12-04
[VEN-2180]: VAI Minting Restricted to Prime Token Holders

web3rover:
## Description

Resolves #

## Checklist

### DIFF
--- a/contracts/Tokens/VAI/VAIController.sol
+++ b/contracts/Tokens/VAI/VAIController.sol
@@ -408,7 +408,7 @@ contract VAIController is VAIControllerStorageG3, VAIControllerErrorReporter, Ex
      * @return uint 0=success, otherwise a failure (see ErrorReporter.sol for details)
      */
     function toggleOnlyPrimeHolderMint() external returns (uint) {
-        _ensureAllowed("toggleOnlyPrimeHolderMint(bool)");
+        _ensureAllowed("toggleOnlyPrimeHolderMint()");
 
         if (!mintEnabledOnlyForPrimeHolder && prime == address(0)) {
             return uint(Error.REJECTION);


contracts/Tokens/VAI/VAIController.sol
```diff
@@ -8,19 +8,26 @@ import "@venusprotocol/governance-contracts/contracts/Governance/IAccessControlM
...
     /// @notice Initial index used in interest computations
     uint public constant INITIAL_VAI_MINT_INDEX = 1e18;
 
     /// @notice Emitted when Comptroller is changed
     event NewComptroller(ComptrollerInterface oldComptroller, ComptrollerInterface newComptroller);
 
+    /// @notice Emitted when mint for prime holder is changed
+    event MintOnlyForPrimeHolder(bool previousMintEnabledOnlyForPrimeHolder, bool newMintEnabledOnlyForPrimeHolder);
+
+    /// @notice Emiitted when Prime is changed
```
chechu, line 28:
```suggestion
    /// @notice Emitted when Prime is changed
```

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -387,6 +394,38 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
 
+    /**
+     * @notice Set the prime token contract address
+     * @param prime_ The new address of the prime token contract
+     * @return uint 0=success, otherwise a failure (see ErrorReporter.sol for details)
+     */
+    function _setPrimeToken(address prime_) external returns (uint) {
+        if (msg.sender != admin) {
+            return fail(Error.UNAUTHORIZED, FailureInfo.SET_COMPTROLLER_OWNER_CHECK);
+        }
```
chechu, line 405: We could use the modifier `onlyAdmin` instead, right? As we are doing at `setAccessControl`. This way, we wouldn't need to return any value

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -387,6 +394,38 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+
+    /**
+     * @notice Toggle mint only for prime holder
+     * @param _mintEnabledOnlyForPrimeHolder set to true to allow minting only for prime holders
+     * @return uint 0=success, otherwise a failure (see ErrorReporter.sol for details)
+     */
+    function toggleOnlyPrimeHolderMint(bool _mintEnabledOnlyForPrimeHolder) external returns (uint) {
+        if (msg.sender != admin) {
+            return fail(Error.UNAUTHORIZED, FailureInfo.SET_COMPTROLLER_OWNER_CHECK);
+        }
```
chechu, line 421: I would use `_ensureAllowed` instead, to be integrated with the ACM. In the VIP we should authorize Normal, Fast-track and Critical timelock to execute this function). And I wouldn't return any value in this function

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -409,6 +448,16 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
 
     // solhint-disable-next-line code-complexity
     function getMintableVAI(address minter) public view returns (uint, uint) {
+        if (mintEnabledOnlyForPrimeHolder && prime == address(0)) {
+            return (uint(Error.REJECTION), 0);
+        }
```
chechu, line 453: I would move this check to the `toggleOnlyPrimeHolderMint` function, to save gas in the user operations. If the new value for `mintEnabledOnlyForPrimeHolder` is true and `prime == address(0)`, I would revert the TX

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -409,6 +448,16 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+
+        if (mintEnabledOnlyForPrimeHolder && prime != address(0)) {
+            if (!IPrime(prime).isUserPrimeHolder(minter)) {
```
chechu, line 456: If you applied the previous suggested changes, we could simplify this check to this:
```
if (mintEnabledOnlyForPrimeHolder && !IPrime(prime).isUserPrimeHolder(minter)) {
  return (uint(Error.REJECTION), 0);
}
```

contracts/Tokens/VAI/VAIControllerStorage.sol
```diff
@@ -84,3 +84,11 @@ contract VAIControllerStorageG2 is VAIControllerStorageG1 {
     /// @notice Access control manager address
     address public accessControl;
 }
+
+contract VAIControllerStorageG3 is VAIControllerStorageG2 {
+    /// @notice The address of the prime contract
```
chechu, line 89: I would clarify that it can be the zero address

contracts/Tokens/VAI/VAIControllerStorage.sol
```diff
@@ -84,3 +84,11 @@ contract VAIControllerStorageG2 is VAIControllerStorageG1 {
...
+    address public prime;
+
+    /// @notice Tracks if minting is enabled only for prime token holders
```
chechu, line 92: I would specify that it can be `true` only if `prime` is not the zero address (assuming the rest of my comments)

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -387,6 +394,34 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+        prime = prime_;
+
+    }
+
+    /**
+     * @notice Toggle mint only for prime holder
+     * @param _mintEnabledOnlyForPrimeHolder set to true to allow minting only for prime holders
+     * @return uint 0=success, otherwise a failure (see ErrorReporter.sol for details)
+     */
+    function toggleOnlyPrimeHolderMint(bool _mintEnabledOnlyForPrimeHolder) external returns (uint) {
```
chechu, line 412: I would call it `setOnlyPrimeHolderMint` if we are going to accept a parameter, or I would remove the parameter if we are going to keep this name (and I would do this internally: `mintEnabledOnlyForPrimeHolder = !mintEnabledOnlyForPrimeHolder`)

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -387,6 +394,34 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+        _ensureAllowed("toggleOnlyPrimeHolderMint(bool)");
+
+        emit MintOnlyForPrimeHolder(mintEnabledOnlyForPrimeHolder, _mintEnabledOnlyForPrimeHolder);
+        mintEnabledOnlyForPrimeHolder = _mintEnabledOnlyForPrimeHolder;
+
+        if (mintEnabledOnlyForPrimeHolder && prime == address(0)) {
+            return uint(Error.REJECTION);
+        }
```
chechu, line 420: I think we should move this check before the emission of the event to avoid any change in the state

contracts/Tokens/VAI/VAIController.sol
```diff
@@ -387,6 +394,32 @@ contract VAIController is VAIControllerStorageG2, VAIControllerErrorReporter, Ex
...
+        emit NewPrime(prime, prime_);
+        prime = prime_;
+    }
+
+    /**
+     * @notice Toggle mint only for prime holder
+     * @return uint 0=success, otherwise a failure (see ErrorReporter.sol for details)
+     */
+    function toggleOnlyPrimeHolderMint() external returns (uint) {
+        _ensureAllowed("toggleOnlyPrimeHolderMint(bool)");
```
chechu, line 411:
```suggestion
        _ensureAllowed("toggleOnlyPrimeHolderMint()");
```

chechu (approved):

PR #405, opened 2023-12-05
[VEN-2051]: Time Manager Integration

web3rover:
## Description

Resolves #

## Checklist

### DIFF
--- a/contracts/Tokens/Prime/IPrime.sol
+++ b/contracts/Tokens/Prime/IPrime.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: BSD-3-Clause
 pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
 
 /**
  * @title IPrime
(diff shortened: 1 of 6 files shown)


contracts/Utils/Constants.sol
chechu, line 1: I would use this constant.sol file: https://github.com/VenusProtocol/solidity-utilities/blob/develop/contracts/constants.sol

We should add the `SECONDS_PER_YEAR` constant, release a new version and start using it here

contracts/Tokens/Prime/PrimeLiquidityProvider.sol
```diff
@@ -95,8 +97,13 @@ contract PrimeLiquidityProvider is
...
     }
 
+    /**
+     * @notice Prime Liquidity Provider constructor
+     * @param _blocksPerYear total blocks per year
+     * @param _timeBased A boolean indicating whether the contract is based on time or block.
+     */
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor() {
+    constructor(uint256 _blocksPerYear, bool _timeBased) TimeManager(_timeBased, _blocksPerYear) {
```
chechu, line 106: We could define the parameters of the constructor in the same order of the parameters in TimeManager: `bool _timeBased, uint256 _blocksPerYear`

chechu (approved):

PR #414, opened 2023-12-18
[VEN -2250] Reduce reserves with available cash

coreyar:
## Description

When accruing interest we reduce reserves. If there isn't enough available cash the accrual would error which  causes problems like preventing fresh account balances from being calculated.

This change reduces the reserves with available cash if it is less then the accrued revenue.


contracts/Tokens/VTokens/VToken.sol
```diff
@@ -508,13 +509,12 @@ contract VToken is VTokenInterface, Exponential, TokenErrorReporter {
         accrualBlockNumber = currentBlockNumber;
         borrowIndex = borrowIndexNew;
         totalBorrows = totalBorrowsNew;
-        totalReserves = totalReservesNew;
```
chechu, line 511: Why are we removing this? If we don't do this we won't increase the reserves every time we accrue interests

contracts/Tokens/VTokens/VToken.sol
```diff
@@ -500,6 +500,7 @@ contract VToken is VTokenInterface, Exponential, TokenErrorReporter {
                 );
         }
 
+        uint availableReserves = totalReservesNew < cashPrior ? totalReservesNew : cashPrior;
```
chechu, line 503: We could move this to the `if` statement where the var `availableReserves` is used. This way we would save some gas most of the cases

contracts/Tokens/VTokens/VToken.sol
```diff
@@ -1496,11 +1496,6 @@ contract VToken is VTokenInterface, Exponential, TokenErrorReporter {
             return fail(Error.MARKET_NOT_FRESH, FailureInfo.REDUCE_RESERVES_FRESH_CHECK);
         }
 
-        // Fail gracefully if protocol has insufficient underlying cash
-        if (getCashPrior() < reduceAmount) {
-            return fail(Error.TOKEN_INSUFFICIENT_CASH, FailureInfo.REDUCE_RESERVES_CASH_NOT_AVAILABLE);
-        }
-
```
chechu, line 1503: I would keep this check. `_reduceReservesFresh` can be invoked "externally" from `_reduceReserves`, where no check compared with the available cash is done

contracts/test/MockVBNB.sol
```diff
@@ -361,7 +358,7 @@ contract MockVBNB is VToken {
         accrualBlockNumber = currentBlockNumber;
         borrowIndex = borrowIndexNew;
         totalBorrows = totalBorrowsNew;
-        totalReserves = totalReservesNew;
+        totalReserves = availableReserves;
```
chechu, line 361: I would say reserves have to be increased always with `totalReservesNew`, independently of having enough cash to cover it

contracts/Tokens/VTokens/VToken.sol
coreyar: Because the updated totalReservesValue is also set in `_ReduceReservesFresh` it seemed logical to calculate it in one place

contracts/Tokens/VTokens/VToken.sol
coreyar: This check is what causes `accrueInterest` to fail. Should we have another one `_reduceFreshReserveSafe` that can be executed in situations outside of `accrueInterest` and includes this check?

chechu: But now we are not invoking `_reduceFreshReserve` with a `reduceAmount` greater than `getCashPrior()`, so `accrueInterest` shouldn't fail anymore

coreyar: That's right

chechu (approved):

chechu (approved):

GitGuru7 (approved):

Debugger022 (approved):

PR #415, opened 2023-12-20
[VEN-2253] vFDUSD Deployment

0xlucian:
## Description

This PR deploys `vFDUSD` market and also adds deployment config and deployment scripts for both `MockTokens` and new markets (similar structure to IL repo)

Resolves #VEN-2253

## Checklist


helpers/deploymentConfig.ts
```diff
@@ -0,0 +1,148 @@
...
+  bscmainnet: DeploymentConfig;
+};
+
+const preconfiguredAddresses = {
+  hardhat: {
+    VTreasury: "account:deployer",
+  },
+  bsctestnet: {
+    VTreasury: venusProtocolBscTestnet.VTreasury,
+    NormalTimelock: governanceBscTestnet.NormalTimelock.address,
```
chechu, line 62: Given we are importing the deployments, shouldn't hardhat help us get these addresses by contract name, instead of directly importing the npm packages?

deploy/010-deploy-mock-tokens copy.ts
GitGuru7, line 1: Rename this file to `deploy/010-deploy-mock-tokens`

GitGuru7 (approved):

helpers/deploymentConfig.ts
coreyar: It is nice to use ethers to get contracts rather than hardcoding a config because it means this will work with the hardhat network. It will also allow use to test the script by running it locally.

helpers/deploymentConfig.ts
```diff
@@ -0,0 +1,174 @@
...
+  bscmainnet: {
+    VTreasury: venusProtocolBscMainnet.VTreasury,
+    NormalTimelock: governanceBscMainnet.NormalTimelock.address,
+    VTokenImpl: "0xc3279442a5acacf0a2ecb015d1cddbb3e0f3f775",
+  },
+};
+
+export const globalConfig: NetworkConfig = {
+  hardhat: {
+    tokensConfig: [],
```
coreyar, line 74: Should we have a similar setup to testnet for hardhat? The idea is that we can use the scripts to easily deploy the protocol for local testing / simulations

deploy/011-deploy-markets.ts
```diff
@@ -0,0 +1,92 @@
...
+  const comptrollerDeployment = await deployments.get("Unitroller");
+
+  console.log(`Got deployment of Unitroller with address: ${comptrollerDeployment.address}`);
+
+  for (const market of marketsConfig) {
+    const { name, asset, symbol, rateModel, baseRatePerYear, multiplierPerYear, jumpMultiplierPerYear, kink_ } = market;
+
+    const token = getTokenConfig(asset, tokensConfig);
+    let tokenContract;
+    if (token.isMock) {
```
coreyar, line 32: Should we also check that the network isn't a test net?

helpers/deploymentConfig.ts
chechu: VEN-2272

chechu (approved):

helpers/deploymentConfig.ts
coreyar: I believe that all of the previous deployments have been imported. Here is the PR - https://github.com/VenusProtocol/venus-protocol/pull/387

deploy/011-deploy-markets.ts
coreyar: I was thinking as an added check because we shouldn't be deploying mocks / underlying tokens on mainnets

helpers/deploymentConfig.ts
chechu: VEN-2296 to track this

PR #429, opened 2024-01-19
[VEN-2322] Unlist Market

web3rover:
## Description

Resolves #

## Checklist


contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -112,6 +130,53 @@ contract MarketFacet is IMarketFacet, FacetBase {
...
+        }
+
+        _actionPaused[market][uint256(Action.BORROW)] = true;
+        _actionPaused[market][uint256(Action.MINT)] = true;
+        _actionPaused[market][uint256(Action.REDEEM)] = true;
+        _actionPaused[market][uint256(Action.REPAY)] = true;
+        _actionPaused[market][uint256(Action.ENTER_MARKET)] = true;
+        _actionPaused[market][uint256(Action.LIQUIDATE)] = true;
+
+        borrowCaps[market] = 0;
```
GitGuru7, line 155: borrowCaps = 0 means unlimited borrowing, see [here](https://github.com/VenusProtocol/venus-protocol/blob/develop/contracts/Comptroller/Diamond/facets/PolicyFacet.sol#L132)

contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -33,7 +33,25 @@ contract MarketFacet is IMarketFacet, FacetBase {
...
+        for (uint256 i = 0; i < accountAssets[account].length; i++) {
+            Market storage market = markets[address(accountAssets[account][i])];
+            if (market.isListed) {
+                assetsIn[index] = accountAssets[account][i];
+                index++;
+            }
+        }
+
+        return assetsIn;
     }
```
Debugger022, line 55: I think we can optimise this piece of code as below

function getAssetsIn(address account) external view returns (VToken[] memory) {

        uint256 len;
        VToken[] storage accountAssetsArray = accountAssets[account];
        uint256 accountAssetsArrayLength = accountAssetsArray.length;

        VToken[] memory assetsIn = new VToken[](accountAssetsArrayLength);

        for (uint256 i; i < accountAssetsArrayLength; ++i) {
            Market storage market = markets[address(accountAssetsArray[i])];
            if (market.isListed) {
                assetsIn[len] = accountAssetsArray[i];
                ++len;
            }
        }

        assembly {
            mstore(assetsIn, len)
        }

        return assetsIn;
}

Debugger022 (approved):

contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -112,6 +130,53 @@ contract MarketFacet is IMarketFacet, FacetBase {
...
+        if (!_market.isListed) {
+            return fail(Error.MARKET_NOT_LISTED, FailureInfo.UNLIST_MARKET_NOT_LISTED);
+        }
+
+        _actionPaused[market][uint256(Action.BORROW)] = true;
+        _actionPaused[market][uint256(Action.MINT)] = true;
+        _actionPaused[market][uint256(Action.REDEEM)] = true;
+        _actionPaused[market][uint256(Action.REPAY)] = true;
+        _actionPaused[market][uint256(Action.ENTER_MARKET)] = true;
+        _actionPaused[market][uint256(Action.LIQUIDATE)] = true;
```
chechu, line 153: I think we could delegate on the other admin actions to perform these changes, and in this function only require the state we want. So, I wouldn't:
* pause the actions
* set borrow and supply caps to zero
* set CF to zero

I would *require* instead:
* every action is paused
* borrow and supply caps are zero
* CF is zero

This way, to unlist a market in a VIP, we could add the following commands:
* pause every action
* set borrow and supply cap to zero
* set CF to zero
* invoke unlistMarket

This would allow `unlistMarket` to be smaller, and easier to test and maintain. At the end, if the "requires" are successful, this function will only set to false the `isListed` attribute and emit the `MarketUnlisted` event.

chechu: Another consequence of this: we wouldn't move to `FacetBase` some of the events we currently have in `SetterFacet`

contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -112,6 +133,42 @@ contract MarketFacet is IMarketFacet, FacetBase {
         return results;
     }
 
+    /**
+     * @notice Unlist a market by setting isListed to false
+     * @dev Checks if market actions are paused and borrowCap/supplyCap/CF are set to 0
+     * @param market The address of the market (token) to unlist
```
Debugger022, line 139:
```suggestion
     * @param market The address of the market (vToken) to unlist
```

contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -33,7 +36,25 @@ contract MarketFacet is IMarketFacet, FacetBase {
...
     function getAssetsIn(address account) external view returns (VToken[] memory) {
-        return accountAssets[account];
+        uint256 len;
+        VToken[] storage _accountAssets = accountAssets[account];
+        uint256 _accountAssetsLength = _accountAssets.length;
+
+        VToken[] memory assetsIn = new VToken[](_accountAssetsLength);
+
+        for (uint256 i; i < _accountAssetsLength; ++i) {
+            Market storage market = markets[address(_accountAssets[i])];
```
Debugger022, line 46:
```suggestion
        VToken[] memory _accountAssets = accountAssets[account];
        uint256 _accountAssetsLength = _accountAssets.length;

        VToken[] memory assetsIn = new VToken[](_accountAssetsLength);

        for (uint256 i; i < _accountAssetsLength; ++i) {
            Market memory market = markets[address(_accountAssets[i])];
```

We can use memory.

deploy/010-facet-upgrade.ts
Debugger022, line 1: We can add the deployment script for other facets too in this same file and use the ones we need.

chechu:
`MarketFacet` is only needed on bsctestnet and bscmainnet. We don't have a Core Comptroller with a Diamond on any other network, so those deployment files should be removed

chechu (approved):

PR #481, opened 2024-05-08
fix: remove duplicate scripts

coreyar:
## Description

Remove duplicate XVS Vault scripts and correctly deploy a single vault


deploy/008-deploy-vaults.ts
```diff
@@ -6,8 +6,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { deploy } = deployments;
   const { deployer } = await getNamedAccounts();
 
-  await deploy("XVSVaultProxy_Implementation", {
-    contract: "XVSVault",
+  await deploy("XVSVault", {
```
Debugger022, line 9: I see that you are changing the implementation contract name here. I think you should also change the name of the deployment file [here](https://github.com/VenusProtocol/venus-protocol/blob/develop/deployments/arbitrumsepolia/XVSVaultProxy_Implementation.json). Because when i deployed `XVSVault` on arbitrum sepolia, I used that contract name and on changing it's name and fetching it's deployment in scripts or anywhere by `XVSVault` name, it will not be able to capture it.

deploy/009-configure-vaults.ts
Debugger022, line 1: I think we can keep the same tags in both deployment and configuration script so that one executes after another which will keep it simple.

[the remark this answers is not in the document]
coreyar: ahh yeah I also realized this and pushed

Debugger022 (approved):

PR #493, opened 2024-06-24
feat: add second mock token

coreyar:
## Description

Adds a second test token when deploying to hardhat


helpers/deploymentConfig.ts
```diff
@@ -12,14 +11,22 @@ export enum InterestRateModels {
...
+      faucetInitialLiquidity?: boolean;
+    }
+  | {
+      isMock: false;
+      name?: string;
+      symbol: string;
+      decimals?: number;
+      tokenAddress: string;
+      faucetInitialLiquidity?: boolean;
+    };
```
kkirka, line 29: Not sure about formatting but maybe extracting the common fields would make the intent more clear?

```suggestion
  (
    {
      isMock: true;
    }
    | {
      isMock: false;
      tokenAddress: string;
    }
  )
  & {
    name?: string;
    symbol: string;
    decimals?: number;
    faucetInitialLiquidity?: boolean;
  }
```

kkirka (approved):

helpers/deploymentConfig.ts
```diff
@@ -12,14 +11,22 @@ export enum InterestRateModels {
...
-  tokenAddress: string;
-  faucetInitialLiquidity?: boolean;
-};
+export type TokenConfig =
+  | {
+      isMock: true;
+      name?: string;
+      symbol: string;
+      decimals?: number;
+      faucetInitialLiquidity?: boolean;
```
Debugger022, line 20: We can remove or set a fixed boolean value in `faucetInitialLiquidity` as for mainnet it will always be false.

helpers/deploymentConfig.ts
coreyar: clarified

helpers/deploymentConfig.ts
coreyar: removed

PR #494, opened 2024-07-02
[VEN-2663] Two Kinks Interest Rate Model

web3rover:
## Description

Resolves #

## Checklist

### DIFF
--- a/contracts/InterestRateModels/TwoKinksInterestRateModel.sol
+++ b/contracts/InterestRateModels/TwoKinksInterestRateModel.sol
@@ -14,12 +14,12 @@ contract TwoKinksInterestRateModel is InterestRateModelV8 {
     ////////////////////// SLOPE 1 //////////////////////
 
     /**
-     * @notice The multiplier of utilization rate per block that gives the slope 1 of the interest rate
+     * @notice The multiplier of utilization rate per block that gives the slope 1 of the interest rate scaled by EXP_SCALE
      */
     int256 public immutable MULTIPLIER_PER_BLOCK;
 
     /**
-     * @notice The base interest rate per block which is the y-intercept when utilization rate is 0
+     * @notice The base interest rate per block which is the y-intercept when utilization rate is 0 scaled by EXP_SCALE
      */
     int256 public immutable BASE_RATE_PER_BLOCK;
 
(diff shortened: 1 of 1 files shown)


contracts/InterestRateModels/TwoKinksInterestRateModel.sol
```diff
@@ -0,0 +1,164 @@
+pragma solidity ^0.5.16;
```
kkirka, line 1: I'm sorry I bring this up a bit late, but do we really have to go with 0.5 here? We could create a separate interface for v8 InterestRateModel and simplify things a bit, both in terms of gas costs (thanks to immutables) and SignedSafeMath (which, AFAIU, has to be audited separately and likely affects the audit costs)

chechu: I would prefer to use 0.8 if it's possible, for sure.

I suppose we would need a new abstract contract `InterestRateModelV8` and extend it here

contracts/Utils/SignedSafeMath.sol
chechu, line 1: We can remove this file, not used anymore

contracts/InterestRateModels/TwoKinksInterestRateModel.sol
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * @notice Calculates the current borrow rate per slot (block), with the error code expected by the market
+     * @param cash The amount of cash in the market
+     * @param borrows The amount of borrows in the market
+     * @param reserves The amount of reserves in the market
+     * @return The borrow rate percentage per slot (block) as a mantissa (scaled by EXP_SCALE)
+     */
+    function _getBorrowRate(uint256 cash, uint256 borrows, uint256 reserves) internal view returns (uint256) {
+        int256 util = int256(utilizationRate(cash, borrows, reserves));
```
chechu, line 150: I suppose it would be unlikely, but shouldn't we check here if the cast is safe? what would happen if the utilization rate is greater than the maximum int256?

kkirka: Will it though? I see there are some restrictions on casting, but no runtime checks (https://docs.soliditylang.org/en/latest/080-breaking-changes.html). The issue about runtime checks is still open: https://github.com/ethereum/solidity/issues/11284

contracts/InterestRateModels/TwoKinksInterestRateModel.sol
```diff
@@ -0,0 +1,178 @@
...
+        uint256 borrowRate = _getBorrowRate(cash, borrows, reserves);
+        uint256 rateToPool = (borrowRate * oneMinusReserveFactor) / EXP_SCALE;
+        return (utilizationRate(cash, borrows, reserves) * rateToPool) / EXP_SCALE;
+    }
+
+    /**
+     * @notice Calculates the utilization rate of the market: `borrows / (cash + borrows - reserves)`
+     * @param cash The amount of cash in the market
+     * @param borrows The amount of borrows in the market
+     * @param reserves The amount of reserves in the market (currently unused)
```
kkirka, line 130:
```suggestion
     * @param reserves The amount of reserves in the market
```

chechu (approved):

PR #498, opened 2024-07-16
VAI Controller in deployment scripts

coreyar:
## Description

- Creates generalized skip functions for skipping all remotes or all source networks
- Fixes VaiController to de deployed on hardhat
- Removes skip from XVS ( it is deployed on all networks)
- Deploy Vai Unitroller and set implementation


deploy/001-comptroller.ts
```diff
@@ -12,7 +14,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const deployerSigner = await hre.ethers.getSigner(deployer);
 
   if (!network.live) {
-    const accessControlManagerDeployment = await deploy("AccessControlManager", {
+    const accessControlManagerDeployment = await deploy("AccessControlManagerV8", {
```
chechu, line 17: I don't find the `AccessControlManagerV8` deployment

coreyar: fixed

deployments/bscmainnet/.migrations.json
```diff
@@ -3,5 +3,6 @@
   "vbnbadmin_deploy": 1697477430,
   "xvs_vault_configuration": 1697032901,
   "configure_prime": 1697032901,
-  "configure_xvs_vaults": 1697032901
+  "configure_xvs_vaults": 1697032901,
+  "vai_controller_config": true
```
chechu, line 7: Is `true` enough? Don't we need a block number?

coreyar: Yes it checks for a truthy value - https://github.com/wighawag/hardhat-deploy/blob/e0ffcf9e7dc92b246e832c4d175f9dbd8b6df14d/src/DeploymentsManager.ts#L1186

But we could also put a block number for completeness

chechu (approved):

PR #499, opened 2024-07-17
refactor: add v1 version of comptroller and vToken for event signatures

coreyar:
## Description

Add legacy version of VBEP20 Token with legacy event signatures


contracts/Tokens/VTokens/legacy/VToken.sol
chechu, line 1: Should we rename the file to `VTokenR1.sol`?

contracts/Tokens/VTokens/legacy/VTokenInterface.sol
chechu, line 1: Should we rename the file to `VTokenInterfaceR1`?

chechu (approved):

PR #505, opened 2024-07-23
refactor: improve token deployments to work on hardhat

coreyar:
## Description

Refactor token deployment scripts to be more compatible with hardhat


deploy/011-deploy-markets.ts
```diff
@@ -78,8 +80,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
         name,
         symbol,
         VTOKEN_DECIMALS,
-        preconfiguredAddresses.NormalTimelock,
-        preconfiguredAddresses.VTokenImpl,
+        network.live ? normalTimelock.address : deployer,
+        vBep20DelegateDeployment.address,
```
chechu, line 84: Nowadays, for `bscmainnet`, the preconfigured address for the VTokenImpl is `0xc3279442a5acacf0a2ecb015d1cddbb3e0f3f775` (set in the `deploymentConfig.ts` file). But, on the deployment files (including `bscmainnet_address.json`), the address of `VBep20Delegate` is `0x6E5cFf66C7b671fA1D5782866D80BD15955d79F6`.

AFAIU, we shouldn't deploy a new `vBep20DelegateDeployment` contract for live networks. Alternatively, we could update the deployment json files, setting the current addresses for the vToken implementations (on bscmainnet and bsctestnet).

Otherwise, I think next time we use this script to deploy a new vToken to a live network, we'll use a wrong implementation address

coreyar: So this was confusing for me My expectation was not to find an implementation listed for each VToken but instead find `one VBep20Delegate` listed
The `${token}_Implementation` address (0xc3279442a5acacf0a2ecb015d1cddbb3e0f3f775) is incorrect. The current implementation is the `VBep20Delegate` 0x6E5cFf66C7b671fA1D5782866D80BD15955d79F6.

I'm not sure how this happened but if I had to guess the deployments didn't follow the scripts and the `VBep20Delegate` delegate was imported.

chechu: I think the `${token}_Implementation` entries were included in the PR https://github.com/VenusProtocol/venus-protocol/pull/387. They are not managed by any deployment script. Maybe we could remove them from the json deployment script

chechu (approved):

PR #524, opened 2024-10-07
[VEN-2846]: Add SolvBTC market

kkirka:
This PR adds BNB testnet & mainnet deployments for vSolvBTC, the corresponding interest rate model, and the mock token for SolvBTC on BNB testnet.


deploy/007-deploy-mock-tokens.ts
```diff
@@ -26,7 +26,5 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
 };
 
 func.tags = ["MockTokens"];
-// These are mock tokens
-func.skip = async hre => hre.network.name !== "hardhat";
```
chechu, line 30: Will not we deploy the mock tokens to mainnet with this change?

helpers/deploymentConfig.ts
```diff
@@ -183,46 +215,27 @@ export const getGlobalConfig: () => Promise<NetworkConfig> = async () => {
...
           collateralFactor: convertToUnit("0.75", 18),
-          liquidationThreshold: convertToUnit("0.8", 18),
-          reserveFactor: convertToUnit("0.1", 18),
-          initialSupply: convertToUnit(9000, 18),
-          supplyCap: convertToUnit(5_500_000, 18),
-          borrowCap: convertToUnit(4_400_000, 18),
-          vTokenReceiver: vTreasuryAddress,
-        },
-        {
-          name: "Venus DOGE",
```
chechu, line 194: Why are we removing vDOGE and vUSDT?

deploy/007-deploy-mock-tokens.ts
kkirka: Tokens are defined in `tokenConfig` mapping, `isMock` should be false for mainnet networks

helpers/deploymentConfig.ts
kkirka: Because they don't have the corresponding `tokenConfig` for these tokens. The alternative would be to specify both token configs and market configs for all the markets.

chechu (approved):

deployments/bsctestnet_addresses.json
```diff
@@ -52,7 +54,7 @@
     "VBNBAdmin": "0x04109575c1dbB4ac2e59e60c783800ea10441BBe",
     "VBNBAdmin_Implementation": "0x90c891bb2b1821ADE159ECa6FbA0418b2bD1b86D",
     "VBNBAdmin_Proxy": "0x04109575c1dbB4ac2e59e60c783800ea10441BBe",
-    "VBep20Delegate": "0xad6aa8Bb4829560412A94AA930745f407BF8000B",
```
GitGuru7, line 55: VBep20Delegate has been redeployed, I think this is not needed.

kkirka: I redeployed the mainnet version of vSolvBTC using the old implementation, and I kept the testnet deployment as is because the market is already added. I will make a testnet VIP changing the implementation to the existing one.

GitGuru7 (approved):

chechu (approved):

PR #543, opened 2024-12-30
Add fork deployments

coreyar:
## Description

Fixes some issues with fork tests
Adds support for forked deployments

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -9,7 +9,7 @@ import "@typechain/hardhat";
 import fs from "fs";
 import "hardhat-dependency-compiler";
 import "hardhat-deploy";
-import { HardhatUserConfig, extendConfig, task } from "hardhat/config";
+import { HardhatUserConfig, extendConfig, extendEnvironment, task } from "hardhat/config";
 import { HardhatConfig } from "hardhat/types";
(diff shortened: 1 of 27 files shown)


hardhat.config.zksync.ts
```diff
@@ -147,7 +147,7 @@ const config: HardhatUserConfig = {
 };
 
 function isFork() {
-  return process.env.FORK === "true"
+  return process.env.FORKED_NETWORK
```
chechu, line 150:
```suggestion
  return !!process.env.FORKED_NETWORK
```

hardhat.config.ts
```diff
@@ -381,14 +406,12 @@ const config: HardhatUserConfig = {
 };
 
 function isFork() {
-  return process.env.FORK === "true"
+  return process.env.FORKED_NETWORK
```
chechu, line 409:
```suggestion
  return !!process.env.FORKED_NETWORK
```

chechu (approved):

hardhat.config.zksync.ts
coreyar: I had this but the linter was giving errors and since something that is truthy (a nonempty string) works I didn't dig deeper into the lint rules

PR #587, opened 2025-04-21
[VEN-3163]: Add checkpoint rate models

kkirka:
This PR adds deployments for [CheckpointView contracts](https://github.com/VenusProtocol/venus-protocol/pull/576) pointing to different rate models – ones with the old block rate before the specified timestamp, and ones with the new block rate after the specified timestamp.

* Old rate models were deployed previously, https://github.com/VenusProtocol/venus-protocol/pull/578 added the old blocks per year suffixes. Note that the "old rate model" is not necessarily equal to the currently configured rate model since duplicates were removed. However, they should be identical in terms of the parameters.
* New rate models (with the reduced block time) were deployed in https://github.com/VenusProtocol/venus-protocol/pull/586

This PR contains two deployment scripts – one that deploys the CheckpointView contracts and __generates a setter contract,__ and one that deploys this setter contract (for each network). These setter contracts will be used in the VIP to configure the new rate models.


chechu (approved):

deploy/17-deploy-checkpoint-setter.ts
```diff
@@ -0,0 +1,27 @@
...
+    contract: getSetterContractName(chain),
+    from: deployer,
+    args: [],
+    log: true,
+    skipIfAlreadyDeployed: true,
+  });
+};
+
+func.tags = ["RateModelCheckpointSetter"];
+func.skip = async hre => !["bsctestnet", "bscmainnet"].includes(hre.network.name);
```
GitGuru7, line 25: Can we use here `skipRemoteNetworks()` function

GitGuru7 (approved):

deploy/17-deploy-checkpoint-setter.ts
kkirka: Uses `skipRemoteNetworks()` now: https://github.com/VenusProtocol/venus-protocol/pull/587/files#diff-e277da072a82c37fdf321620435c07718a00b72bfa254623135f18fb2341a941R26

kkirka: I changed it initially but I reverted the change since "non-remote networks" include hardhat, and we don't want to execute this script on hardhat network

chechu (approved):

PR #608, opened 2025-07-28
[VEN-3343] : Whitelist Flash-loan Users

Debugger022:
## Description

Add a new check on the Flash loan feature, to authorize the execution of the flash loans only to some whitelisted accounts. We should:
- add a new storage variable with the authorized accounts
- setter function, controlled by the ACM, to add and remove accounts from the whitelist
- add new check in the flashLoan functions

### DIFF
--- a/contracts/Comptroller/Diamond/facets/PolicyFacet.sol
+++ b/contracts/Comptroller/Diamond/facets/PolicyFacet.sol
@@ -386,6 +386,7 @@ contract PolicyFacet is IPolicyFacet, XVSRewardsHelper {
      *      - Reverts with `Insufficient reypayment balance` if the repayment (amount + fee) is insufficient after the operation.
      */
     function executeFlashLoan(
+        address initiator,
         address payable receiver,
         VToken[] calldata vTokens,
         uint256[] calldata underlyingAmounts,
@@ -395,6 +396,10 @@ contract PolicyFacet is IPolicyFacet, XVSRewardsHelper {
         // Asset and amount length must be equals and not be zero
         if (vTokens.length != underlyingAmounts.length || vTokens.length == 0) {
             revert("Invalid flashLoan params");
+        }
+
+        if (!authorizedFlashLoan[initiator]) {
+            revert("Flash loan not authorized for this account");
         }
 
         uint256[] memory protocolFees = new uint256[](vTokens.length);
--- a/contracts/Comptroller/Diamond/facets/SetterFacet.sol
+++ b/contracts/Comptroller/Diamond/facets/SetterFacet.sol
@@ -93,6 +93,9 @@ contract SetterFacet is ISetterFacet, FacetBase {
     /// @notice Emitted when XVS vToken address is changed
     event NewXVSVToken(address indexed oldXVSVToken, address indexed newXVSVToken);
 
+    /// @notice Event emitted when an account is added to flash loan whitelist
+    event IsAccountFlashLoanWhitlisted(address indexed account, bool indexed isWhitelisted);
+
     /**
      * @notice Compare two addresses to ensure they are different
      * @param oldAddress The original address to compare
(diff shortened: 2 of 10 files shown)


contracts/Comptroller/Diamond/facets/SetterFacet.sol
```diff
@@ -603,4 +606,17 @@ contract SetterFacet is ISetterFacet, FacetBase {
         emit NewXVSVToken(xvsVToken, xvsVToken_);
         xvsVToken = xvsVToken_;
     }
+
+    /**
+     * @notice Adds/Removes an account to the flash loan whitelist
+     * @param account The account to authorize for flash loans
+     * @param _isWhiteListed True to whitelist the account for flash loans, false to remove from whitelist
+     */
+    function _setWhiteListFlashLoanAccount(address account, bool _isWhiteListed) external {
```
chechu: I think it's ok if we stop using underscore for external functions in the Core pool on BNB Chain

contracts/Comptroller/Diamond/facets/PolicyFacet.sol
```diff
@@ -386,6 +386,7 @@ contract PolicyFacet is IPolicyFacet, XVSRewardsHelper {
      *      - Reverts with `Insufficient reypayment balance` if the repayment (amount + fee) is insufficient after the operation.
      */
     function executeFlashLoan(
+        address initiator,
```
chechu: Do we need this? Shouldn't we check if `msg.sender` is whitelisted instead?

contracts/Comptroller/Diamond/facets/SetterFacet.sol
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/608/commits/13db780b87f565b67214fbe459691aac59ac5af9

contracts/Comptroller/Diamond/facets/PolicyFacet.sol
Debugger022: `executeFlashloan` needs to be called inside another contract in order to receive flashloan . Using `msg.sender` inside `executeFlashloan` will refer to the immediate caller which is contract address and not the user who started the Transaction .

contracts/Comptroller/Diamond/facets/SetterFacet.sol
```diff
@@ -603,4 +606,17 @@ contract SetterFacet is ISetterFacet, FacetBase {
...
         xvsVToken = xvsVToken_;
     }
+
+    /**
+     * @notice Adds/Removes an account to the flash loan whitelist
+     * @param account The account to authorize for flash loans
+     * @param _isWhiteListed True to whitelist the account for flash loans, false to remove from whitelist
+     */
+    function setWhiteListFlashLoanAccount(address account, bool _isWhiteListed) external {
+        ensureAllowed("setWhiteListFlashLoanAccount(address)");
```
chechu:
```suggestion
        ensureAllowed("setWhiteListFlashLoanAccount(address,bool)");
```

Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/608/commits/8192bab1d78f14cd5c2d2c2a929bfdff075a2699

contracts/Tokens/VTokens/VToken.sol
```diff
@@ -418,19 +419,24 @@ abstract contract VToken is VTokenInterface, Exponential, TokenErrorReporter {
...
         uint256 amount,
         bytes calldata param
     ) external nonReentrant returns (uint256) {
-        if (!isFlashLoanEnabled) revert("FlashLoan not enabled");
+        if (!isFlashLoanEnabled) revert FlashLoanNotEnabled();
         ensureNonZeroAddress(receiver);
 
+        // Check if the caller is authorized to execute flash loans
+        if (!comptroller.authorizedFlashLoan(initiator)) revert FlashLoanNotAuthorized();
+
```
GitGuru7, line 441: Validating `authorizedFlashLoan` for the initiator argument alone is not sufficient, since anyone could pass a different address. It’s safer to use msg.sender directly as the initiator and validate that, rather than taking initiator as a function argument.

Debugger022: `executeFlashloan` needs to be called inside another contract in order to receive flashloan . Using msg.sender inside `executeFlashloan` will refer to the immediate caller which is contract address and not the user who started the Transaction .  if we use `msg.sender` instead of `initiator`, then we validate the contract who receives the flashloan .

PR #629, opened 2025-09-24
Update facet-cut-params-generator to handle add and replace selectors dynamically

GitGuru7:
## Description
This PR updates the `facet-cut-params-generator` script to improve how diamond cut parameters are generated:

* The script now **fetches the current function selectors** from the deployed diamond contract.
* It generates the **final cut-params array**, considering both **Add** and **Replace** operations dynamically.
* For `IFacetBase` selectors:

  * If there are **no new selectors**, the resulting cut-params JSON can be **used directly**.
  * If there **are new selectors**, they need to be **manually assigned** to the appropriate facet addresses before using the cut-params.
  
  **Note**: Now onwards when running the script, please use the` --network` flag to specify the target network.

### DIFF
--- a/script/deploy/comptroller/facet-cut-params-generator.ts
+++ b/script/deploy/comptroller/facet-cut-params-generator.ts
@@ -1,5 +1,5 @@
 import fs from "fs";
-import { ethers } from "hardhat";
+import { deployments, ethers, network } from "hardhat";
 
 import { FacetCutAction, getSelectors } from "./diamond";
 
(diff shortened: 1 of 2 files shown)


script/deploy/comptroller/facet-cut-params-generator.ts
```diff
@@ -31,73 +23,72 @@ const FacetsInterfaces = {
...
-        break;
-      }
+  const cut: any[] = [];
+
+  for (let i = 0; i < FacetNames.length; i++) {
+    const facetName = FacetNames[i];
+    const newFacetAddress = newFacetAddresses[facetName];
+
+    // Current facet selectors (from diamond, assuming the facet addresses in the same order as newFacetAddresses)
+    const currentSelectors: string[] = facetName === "FacetBase" ? [] : currentFacets[i].functionSelectors;
```
Debugger022, line 56: What if the diamond's facet order differs from `FacetNames`?

script/deploy/comptroller/facet-cut-params-generator.ts
```diff
@@ -31,73 +23,72 @@ const FacetsInterfaces = {
...
+
+    // New facet selectors (from interface)
+    const newFacetInterface = await ethers.getContractAt(FacetsInterfaces[facetName], newFacetAddress);
+    const newSelectors: string[] = getSelectors(newFacetInterface);
```
Debugger022, line 60: Are there any chances that `getSelectors` could fail?

PR #630, opened 2025-09-26
[VPD-150]: Add PT-USDe-30Oct2025 market on BNB Chain Core pool 

GitGuru7:
## Description

Resolves #

## Checklist


helpers/markets/bsctestnet.ts
```diff
@@ -634,4 +634,26 @@ export default [
       vTokenReceiver: "VTreasury",
     },
   },
+  {
+    name: "Venus PT-USDe-30Oct2025",
```
chechu: We should use `OCT` instead of `Oct` for mainnet. Not needed to change it on testnet

kkirka (approved):

PR #642, opened 2025-10-23
[VPD-226] Flash Loan Quantstamp Audit Mitigations

Debugger022:


web3rover (approved):

contracts/Comptroller/Diamond/facets/FlashLoanFacet.sol
```diff
@@ -306,4 +321,12 @@ contract FlashLoanFacet is IFlashLoanFacet, FacetBase, ReentrancyGuardTransient
...
         );
     }
+
+    /**
+     * @notice Check if flash loans are currently paused system-wide
+     * @return True if flash loans are paused, false otherwise
+     */
+    function isFlashLoanPaused() external view returns (bool) {
+        return flashLoanPaused;
+    }
```
GitGuru7: Do we need this explicit getter? Since `flashLoanPaused` is **public**, it automatically exposes a getter. We could just add `flashLoanPaused() external view returns (bool); `in the ComptrollerInterface instead.

Debugger022: removed ec80cec473793415abd737d336f6e1af984c2764

contracts/Comptroller/Diamond/facets/FlashLoanFacet.sol
```diff
@@ -18,6 +18,9 @@ import { ReentrancyGuardTransient } from "../../../Utils/ReentrancyGuardTransien
  *      The contract supports protocol fee collection and integrates with the Venus lending protocol.
  */
 contract FlashLoanFacet is IFlashLoanFacet, FacetBase, ReentrancyGuardTransient {
+    /// @notice Maximum number of assets that can be flash loaned in a single transaction
+    uint256 public constant MAX_FLASHLOAN_ASSETS = 200;
```
fred-venus, line 22: curious how did u actually decide the value here, i noticed u started with 10 and ended up with 200

Debugger022: > curious how did u actually decide the value here, i noticed u started with 10 and ended up with 200

It's currently near the upper threshold. We can adjust it if needed, but keeping it at `200` is safer for now.

fred-venus (approved):

PR #650, opened 2025-11-13
Unlist Market from eMode Pools

web3rover:
## Description

Resolves #

## Checklist

### DIFF
--- a/contracts/Comptroller/Diamond/facets/MarketFacet.sol
+++ b/contracts/Comptroller/Diamond/facets/MarketFacet.sol
@@ -217,6 +217,18 @@ contract MarketFacet is IMarketFacet, FacetBase {
         require(_market.collateralFactorMantissa == 0, "collateral factor is not 0");
 
         _market.isListed = false;
+
+        for (uint96 i = lastPoolId; i > corePoolId; i--) {
+            address[] memory markets = getPoolVTokens(uint96(i));
+
+            for (uint256 j = 0; j < markets.length; j++) {
+                if (markets[j] == market) {
+                    _removePoolMarket(i, market);
+                    break;
+                }
+            }
+        }
+
         emit MarketUnlisted(market);
 
         return uint256(Error.NO_ERROR);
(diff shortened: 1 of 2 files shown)


contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -217,6 +217,18 @@ contract MarketFacet is IMarketFacet, FacetBase {
         require(_market.collateralFactorMantissa == 0, "collateral factor is not 0");
 
         _market.isListed = false;
+
+        for (uint96 i = lastPoolId; i > corePoolId; i--) {
+            address[] memory markets = getPoolVTokens(uint96(i));
+
+            for (uint256 j = 0; j < markets.length; j++) {
+                if (markets[j] == market) {
```
GitGuru7: I understand this is to avoid a revert in `_removePoolMarket` if the market is not already listed in eMode. I would prefer using a simple check like `if (_poolMarkets[index].isListed) `similar to what we have in `_removePoolMarket` instead of looping over all poolMarkets.

contracts/Comptroller/Diamond/facets/MarketFacet.sol
```diff
@@ -217,6 +217,18 @@ contract MarketFacet is IMarketFacet, FacetBase {
         require(_market.collateralFactorMantissa == 0, "collateral factor is not 0");
 
         _market.isListed = false;
```
GitGuru7: Nit: We currently don’t allow adding a market to a pool that isn’t listed in the Core Pool. It might be better to set `_market.isListed = false` after removing it from all pools (delisting from eMode). Functionally it doesn’t change behavior, but it aligns the logic more clearly.

tests/hardhat/Comptroller/Diamond/assetListTest.ts
```diff
@@ -145,6 +145,7 @@ describe("Comptroller: assetListTest", () => {
 
     const receipt = await comptroller.connect(customer).unlistMarket(unlistToken.address);
     expect(receipt).to.emit(unitroller, "MarketUnlisted");
+    expect(receipt).to.emit(unitroller, "PoolMarketRemoved");
```
GitGuru7, line 154: I see we already added an event check in the test. Could we also include a simple test to verify that an unlisted market is no longer present in any eMode pool ?

GitGuru7 (commented):
Overall LGTM! Left a couple of nitpicks please let me know if they make sense to you.

contracts/Comptroller/Diamond/facets/MarketFacet.sol
fred-venus: agree, let's avoid looping here which could waste ton of gas
```
PoolMarketId.wrap(bytes32((uint256(poolId) << 160) | uint160(vToken)));
```

fred-venus:
Also, pls ask them to audit this part as well

contracts/Comptroller/Diamond/facets/MarketFacet.sol
web3rover: Fixed

contracts/Comptroller/Diamond/facets/MarketFacet.sol
web3rover: Fixed

tests/hardhat/Comptroller/Diamond/assetListTest.ts
web3rover: Fixed

fred-venus:
we will gradually migrate emode/isolation mode to spoke pool

PR #657, opened 2026-01-06
[VPD-418]: Add U market on the BNB Chain Core Pool

Debugger022:
This PR adds the United Stables (U) market on the BNB chain core pool.


fred-venus:
the contract lgtm, but i was thinking should we deploy checkpoint instead of irm directly so that we won't need to include this into the bnb hardfork vip any more (considering that one would be executed later than this)

web3rover (approved):

fred-venus:
checkpoint reviewed lgtm

fred-venus (approved):

PR #668, opened 2026-03-31
[VPD-893]: Integrate DeviationBoundedOracle for conservative CF path

GitGuru7:
## Description
 Integrate DeviationBoundedOracle into Comptroller for conservative CF-path pricing

- Introduce `ComptrollerV19Storage` with a new `deviationBoundedOracle` slot
  and add `setDeviationBoundedOracle` admin setter in `SetterFacet`
- Call `dbo.updateProtectionState()` for all account assets before liquidity
  checks in `redeemAllowed` and `borrowAllowed` (populates transient price cache)
- In `ComptrollerLens`, use bounded collateral/debt prices from the DBO in the
  collateral-factor path; liquidation-threshold path continues to use spot prices

---

⚠️ for tests, please update the oracle package to use the bounded oracle changes from this PR
https://github.com/VenusProtocol/oracle/pull/306

you can either:

* use `npm pack` to create a local package
* or directly link it locally using:

```json id="k9x3d1"
"@venusprotocol/oracle": "link:../oracle"
```


contracts/Lens/ComptrollerLens.sol
```diff
@@ -226,6 +236,11 @@ contract ComptrollerLens is ComptrollerLensInterface, ComptrollerErrorReporter,
         // For each asset the account is in
         VToken[] memory assets = ComptrollerInterface(comptroller).getAssetsIn(account);
         uint assetsCount = assets.length;
+        if (weightingStrategy == WeightFunction.USE_COLLATERAL_FACTOR) {
+            vars.boundedOracle = ComptrollerInterface(comptroller).deviationBoundedOracle();
```
fred-venus, line 240: meaning in this case spotOracle = addr(0) ?  why not we always return both ?

fred-venus:
I guess we will need to update the VenusLens contract as well. It's not directly used by any other contract, but I'm thinking that fe/be might have certain dependencies. Probably we could return three prices instead of just the spot price.

fred-venus:
> I guess we will need to update the VenusLens contract as well. It's not directly used by any other contract, but I'm thinking that fe/be might have certain dependencies. Probably we could return three prices instead of just the spot price.

also same thing to `SnapshotLens`, better confirm with FE/BE if they are reply on it in some way

fred-venus:
in VAI controller#getMintableVAI, should we update this function as well ? What I saw is that we are still using the spot price instead of the updated DeviationBoundedOracle price.


_"Function that returns the amount of VAI a user can mint based on their account liquidy and the VAI mint rate"_

contracts/Lens/ComptrollerLens.sol
GitGuru7: Yes, because only one of the oracle is used based on the chosen weighting strategy.

contracts/Comptroller/Diamond/facets/FacetBase.sol
```diff
@@ -260,27 +300,17 @@ contract FacetBase is IFacetBase, ComptrollerV18Storage, ExponentialNoError, Com
...
-     *                          - `WeightFunction.USE_LIQUIDATION_THRESHOLD` to use liquidation threshold
-     * @return (possible error code (semi-opaque),
-     * account liquidity in excess of collateral requirements,
-     * account shortfall below collateral requirements)
+     * @notice Updates the DeviationBoundedOracle protection state for all assets the account has entered
+     * @dev This populates the transient price cache so subsequent view calls in the same transaction
+     *      are gas-efficient. Should be called before any liquidity calculation in the CF path.
+     * @param account The account whose collateral assets need protection state updates
      */
-    function _getAccountLiquidity(
```
fred-venus, line 272: so this `_getAccountLiquidity` is removed ?

GitGuru7: yes just its just a additional refactor, It was originally just a wrapper around getHypotheticalAccountLiquidity. If we decide to keep it, we would need to maintain two versions, one view and one non-view.

Functionally, it only abstracts setting the hypothetical borrow/mint values to zero in the `getHypotheticalAccountLiquidityInternal`. However, this abstraction is already handled in the main external `getAccountLiquidity` function, which makes the wrapper unnecessory.
And  this also improves readability that there only exist one getHypotheticalAccountLiquidity (and no internal wrapper which does the same thing)

fred-venus:
While I don't think this is a major issue or that it will have any substantial impact, would it be better to use a Deviation Oracle here? As thats what would actually be used in a borrowing scenario.

CleanShot 2026-04-10 at 15 15 36@2x

contracts/Comptroller/Diamond/facets/FacetBase.sol
fred-venus: yeah, agree its fine to remove it

fred-venus:
lgtm, let's fix the ci

fred-venus (approved):

PR #683, opened 2026-06-22
[VPD-1430, 1431, 1432] feat(bstock): atomic backstop liquidator + off-chain scripts 

Debugger022:
## Description

Backstop liquidation for bStock (ERC-8056 tokenized stocks: TSLAB, NVDAB, SPCXB) collateral. bStock is RFQ-only (Native is the sole market maker, no AMM), so third-party liquidators may not show up. This adds Venus's own fallback liquidator plus the off-chain tooling to drive it. Liquidation stays permissionless; this is only the backstop so the protocol never carries bad debt on an RFQ-only asset.

### VPD-1430: `BStockLiquidator` contract + tests

Upgradeable, owner/operator-gated atomic liquidator. In one tx it repays an underwater borrow, seizes + redeems the bStock vToken, and sells the bStock to USDT via a pre-signed Native firm-quote with a hard `minOut` (a bad or late quote reverts the whole tx, so the protocol never holds the RFQ-only asset across a price-drift window).

- Two funding modes: **inventory** (own USDT) and **flash** (Venus `executeFlashLoan`, repaid + premium in the same tx).
- Routes the repay through the pool-wide Venus Liquidator gate when set, else liquidates directly; seize measured by balance delta so the liquidator treasury cut is handled correctly.
- Router-allowlisted and operator-gated (it custodies funds and forwards caller-supplied swap calldata to an external router).
- Extends the shared `IComptroller` (`InterfacesV8`) with `getAccountLiquidity` and `executeFlashLoan` (additive).

### VPD-1431: off-chain liquidation scripts

The on-chain contract only does the atomic settle; every liquidation must be paired with an off-chain Native firm-quote. These scripts are that off-chain half.

### VPD-1432: Safe multisig fallback

A read-only generator for a Safe Transaction Builder batch, used to settle a liquidation manually when the Native RFQ path is unavailable.

## Scripts (`scripts/bstock/`)

**`lib/native.ts`** — Native (nativefi) RFQ API client (library). Orderbook / indicative / firm-quote, parses the signed `txRequest` + TTL, holds the TSLAB/NVDAB/USDT constants. API key via `NATIVE_API_KEY`, never hardcoded. Not run directly.

**`native-quote.ts`** — live quote smoke test (read-only, no chain). Sanity-checks the Native integration and shows price / spread / TTL / book depth before anything goes on-chain (depth is thin, so quotes above book size fail).

```
NATIVE_API_KEY=... npx hardhat run scripts/bstock/native-quote.ts
# options (env): TOKEN=TSLAB|NVDAB (def TSLAB), AMOUNT=1, FROM=0x..
```

**`atomic-liquidate.ts`** — driver for the deployed `BStockLiquidator` (the normal path). Confirms shortfall, precomputes the seize (adjusted for the redeem `treasuryPercent`), fetches the MM-signed firm-quote with the contract as taker, checks the router is allowlisted, computes `minOut`, then calls `liquidate` (inventory) or `flashLiquidate` (flash loan).

```
NATIVE_API_KEY=... LIQUIDATOR=0x.. BORROWER=0x.. VBSTOCK=0x.. VDEBT=0x.. REPAY_AMOUNT=5000 npx hardhat run scripts/bstock/atomic-liquidate.ts --network bscmainnet
# MODE=inventory|flash (def inventory), SLIPPAGE=0.5, MIN_OUT_BUFFER=0.5, DRY_RUN=1 (callStatic only), MOCK_NATIVE=<router>:<calldata> (fork tests)
```

**`lib/safe.ts`** — Safe Transaction Builder JSON helper (library). Encodes calls and builds the batch JSON the Safe{Wallet} Transaction Builder imports. Not run directly.

**`safe-fallback.ts`** — manual multisig fallback (read-only, emits JSON). When the RFQ path is down (API outage, halt, weekend, thin depth), a Safe multisig settles manually. Emits a 4-tx batch (approve, `liquidateBorrow`, `redeem`, transfer raw bStock to a custody/Binance address), seize amount snapshotted from chain. Sends nothing.

```
RPC_URL=... BORROWER=0x.. VBSTOCK=0x.. VDEBT=0x.. REPAY_AMOUNT=5000 TARGET=0x.. npx ts-node scripts/bstock/safe-fallback.ts
# ALLOW_PLACEHOLDER=1 emits a draft with a zero TARGET; OUT=path overrides output
```

## Testing

- **Unit (contract)** `tests/hardhat/BStockLiquidator.ts`: 23 passing — both funding modes, gate + treasury-cut path, all guards and error paths.
- **Driver (mock)** `tests/hardhat/BStockAtomicLiquidate.ts`: 4 passing — drives `atomic-liquidate.ts` end to end (inventory + flash, healthy-borrower and router-allowlist guards).
- **Fork (real Venus Core)** `tests/hardhat/Fork/BStockLiquidatorFork.ts`: passing — real `liquidateBorrow` → seize → redeem through the live Venus Liquidator, with strict profit/borrow assertions (stand-in markets + mock router, since no bStock market is listed yet).
- **Native RFQ client** validated against the live API (quote, TTL, router).

## Notes

- No bStock market is listed in Core yet, so a full real liquidation and live RFQ settlement can only be exercised on mainnet once a market exists. The contract is collateral-agnostic and retargets to vTSLAB/vUSDT via env once listed.
- Flash-loan authorization for the liquidator (whitelist + per-vToken `setFlashLoanEnabled`) is handled by a separate VIP (VPD-1433).


contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+
+    // --------------------------------------------------------------------- //
+    //                          INVENTORY mode                               //
+    // --------------------------------------------------------------------- //
+
+    /// @inheritdoc IBStockLiquidator
+    function liquidate(
+        LiquidationParams calldata p
+    ) external override onlyOperator nonReentrant returns (uint256 usdtOut) {
+        _check(p);
```
GitGuru7: **Nit:** Could we rename `_check` to something more descriptive, such as `_validateLiquidationRequest()` or `_validateLiquidationParams` ? Also, instead of using `p` throughout the contract, consider renaming it to improve readability. Since `LiquidationParams` is used extensively, the more descriptive names make the code easier to follow.

contracts/BStock/IBStockLiquidator.sol
```diff
@@ -0,0 +1,110 @@
...
+/// @notice External API, events, errors and parameter struct for {BStockLiquidator}.
+/// @dev The flash-loan callback `executeOperation` is intentionally NOT part of this interface — it is
+///      declared by {IFlashLoanReceiver} (owned by the Core flash-loan subsystem) and implemented directly
+///      by {BStockLiquidator}, keeping this interface free of the concrete `VToken` dependency.
+interface IBStockLiquidator {
+    /// @notice Parameters for a single liquidation.
+    struct LiquidationParams {
+        address borrower; // account to liquidate
+        IVBep20 vDebt; // borrowed market to repay (e.g. vUSDT)
+        IVBep20 vBStock; // bStock collateral market to seize (e.g. vTSLAB)
```
GitGuru7, line 24: rename to **vBStockCollateral** ?

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+        //    `vToken.liquidateBorrow` reverts UNAUTHORIZED, so we route the repay through that
+        //    Venus Liquidator (it pulls our repay and sends us our share of the seized collateral,
+        //    treasury keeps a cut). If the gate is unset, we liquidate directly. Either way the
+        //    seized amount is read as a BALANCE DELTA, so the Liquidator's cut is handled correctly.
+        uint256 vBefore = p.vBStock.balanceOf(address(this));
+        address gate = comptroller.liquidatorContract();
+        if (gate == address(0)) {
+            debt.forceApprove(address(p.vDebt), p.repayAmount);
+            uint256 errCode = p.vDebt.liquidateBorrow(p.borrower, p.repayAmount, p.vBStock);
+            if (errCode != 0) revert LiquidateBorrowFailed(errCode);
```
GitGuru7: Is this fallback path actually necessary? Would we ever expect the Liquidator contract to be unset on the Comptroller?

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+        VToken[] calldata vTokens,
+        uint256[] calldata amounts,
+        uint256[] calldata premiums,
+        address initiator,
+        address /* onBehalf */,
+        bytes calldata param
+    ) external override returns (bool success, uint256[] memory repayAmounts) {
+        if (msg.sender != address(comptroller)) revert OnlyComptroller();
+        if (!_flashActive) revert NoFlashInFlight();
+        if (initiator != address(this)) revert BadInitiator(initiator);
```
GitGuru7: `if (initiator != address(this)) revert BadInitiator(initiator);` should be sufficient. Could we remove the `_flashActive` check to avoid the extra storage read?

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+    /// @inheritdoc IBStockLiquidator
+    function setRouter(address router, bool allowed) external override onlyOwner {
+        ensureNonzeroAddress(router);
+        isRouter[router] = allowed;
+        emit RouterSet(router, allowed);
+    }
+
+    /// @inheritdoc IBStockLiquidator
+    function sweep(address token, address to, uint256 amount) external override onlyOwner {
+        IERC20Upgradeable(token).safeTransfer(to, amount);
```
GitGuru7, line 209: please add a zero-address check here

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    // --------------------------------------------------------------------- //
+    //                              Core                                     //
+    // --------------------------------------------------------------------- //
+
+    /// @dev Pre-flight: router must be allowlisted and the borrower must have a shortfall.
+    function _check(LiquidationParams calldata p) private view {
+        if (!isRouter[p.router]) revert RouterNotAllowed(p.router);
+        (, , uint256 shortfall) = comptroller.getAccountLiquidity(p.borrower);
```
GitGuru7: The original Liquidator contract also supports force liquidations. This check would break that flow for us here.

scripts/bstock/atomic-liquidate.ts
```diff
@@ -0,0 +1,199 @@
...
+  );
+  if (!seizeErr.eq(0)) throw new Error(`liquidateCalculateSeizeTokens error ${seizeErr}`);
+  const exchangeRate: BigNumber = await vBStock.exchangeRateStored();
+  const ONE = BigNumber.from(10).pow(18);
+  // Core redeem routes `treasuryPercent` of the redeemed underlying to the treasury, so the contract
+  // ends up holding LESS than seizeTokens*rate. The quote must match what we actually hold, else the
+  // Native router pull (fixed amountIn) reverts. 0 today, but governance-settable.
+  const treasuryPercent: BigNumber = await comptroller.treasuryPercent();
+  const seizedRaw = seizeTokens.mul(exchangeRate).div(ONE).mul(ONE.sub(treasuryPercent)).div(ONE);
+  const seizedHuman = ethers.utils.formatUnits(seizedRaw, bStockDec);
```
GitGuru7: We should also consider deducting the protocol's share taken by the Original Liquidator contract. Otherwise, this value will overestimate the amount that the BStockLiquidator contract actually receives.

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+        ensureNonzeroAddress(initialOwner);
+        __Ownable2Step_init();
+        __ReentrancyGuard_init();
+        _transferOwnership(initialOwner);
+    }
+
+    // --------------------------------------------------------------------- //
+    //                               Admin                                   //
+    // --------------------------------------------------------------------- //
+
```
fred-venus, line 169: lets have 0xCCa5a587eBDBe80f23c8610F2e53B03158e62948 i.e. https://app.notion.com/p/EBrake-Emergency-Safe-BSC-342b4e5771bb811fb1c1e641194cdb4f as the owner

fred-venus: and then grant each signer inside as operator

contracts/BStock/BStockLiquidator.sol
fred-venus: for now the value seems to be `https://bscscan.com/address/0x0870793286aaDA55D39CE7f82fb2766e8004cF43#code` so all the onchain liquidation events will fall into else case ?

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+        uint256 seizedV = p.vBStock.balanceOf(address(this)) - vBefore;
+
+        // 2. Redeem the seized vBStock for raw bStock. Measure by DELTA so any pre-existing bStock
+        //    (dust or a stray transfer) is excluded — we only sell what this redeem actually returned.
+        uint256 rawBefore = bStock.balanceOf(address(this));
+        uint256 redeemErr = p.vBStock.redeem(seizedV);
+        if (redeemErr != 0) revert RedeemFailed(redeemErr);
+        seizedBStock = bStock.balanceOf(address(this)) - rawBefore;
+
+        // 3. Sell the bStock to USDT via the allowlisted Native router (MM-signed calldata).
```
fred-venus: 1. logically i dont see we need to force the output token to be debt token or anything if its inventory mode ? 

2. and also i see we name `usdtBefore` but its from `debt.balanceOf(address(this));` which means we are assuming that the swap output will always be debt token instead of USDT -> this should only be compulsory  if its flashlaon mode ?

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,255 @@
...
+        emit Liquidated(p.borrower, address(p.vBStock), p.repayAmount, seizedBStock, usdtOut, true);
+        return (true, repayAmounts);
+    }
+
+    // --------------------------------------------------------------------- //
+    //                              Core                                     //
+    // --------------------------------------------------------------------- //
+
+    /// @dev Pre-flight: router must be allowlisted and the borrower must have a shortfall.
+    function _check(LiquidationParams calldata p) private view {
```
fred-venus: i prefer remove `_check` here:

1. this contract itself wont have any other privilege except allowing to use flashloan 
2. the contract has strict permission control i.e. actions can only be trigged by us 

Based on those 2, i dont think having router whitelist mechanism is necessary and when there is really a PK leak there shouldnt be any material loss ?

contracts/BStock/BStockLiquidator.sol
fred-venus: can we further split the flow here ?  ideally no matter what mode:

1. we will end up with a token (even flashloan mode we will have some liqudation income), maybe a futher function to swap the token to stablecoin
2. we are assuming the contract will always have debt token at the beginning, thats also somewhere we should consider we might need to have a swap ahead

fred-venus: Another way might be to make the sweep function operator-allowed instead of restricted to the owner. This way, the operator can:
1. Sweep the token from this contract
2. Swap it to the target debt token
3. Transfer it back to this contract to continue

the drawback for this flow) is that it is not atomic, meaning the whole sequence of actions won't happen in the same block.

contracts/BStock/BStockLiquidator.sol
Debugger022: Two separate checks here, treating them differently:

1. `Shortfall check` → `removing`. Redundant with Core's `liquidateBorrowAllowed`, and it breaks force liquidations (`isForcedLiquidationEnabled` liquidates healthy accounts where `shortfall == 0`).
2. `Router allowlist` → `keeping`. Not about operator gating — it's the low-level call itself:

`(bool ok, ) = p.router.call(p.swapCalldata);`

`router + swapCalldata` are caller-supplied, so this is an arbitrary call with the contract as `msg.sender`. Without `isRouter`, a leaked operator key can set `router = USDT, swapCalldata = transfer(attacker, balance)` and drain inventory directly — `minOut` doesn't help (guards swap output, not funds moved out). It's one `SLOAD`, and the model we documented in the header. Operator gating stops outsiders; the allowlist bounds a compromised insider.

contracts/BStock/BStockLiquidator.sol
Debugger022: [6846758](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/684675809e716d420fa0104918bc6ed8a52285f4)
`_check` is now `_validateRouter` (it only validates the router after we removed the shortfall pre-check), and `p` → `params` throughout the contract + interface. Since it now needs a single field, it takes `address router` instead of the whole struct.

contracts/BStock/BStockLiquidator.sol
Debugger022: [6846758](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/684675809e716d420fa0104918bc6ed8a52285f4)  removed the shortfall pre-check entirely.

contracts/BStock/BStockLiquidator.sol
Debugger022: Right — the gate is `0x0870793286aaDA55D39CE7f82fb2766e8004cF43` on mainnet, so every liquidation goes through the Venus Liquidator. The `if/else` was only there as a safety/defensive fallback in case the gate was ever unset. Since we can rely on it always being set, I removed the direct fallback branch entirely and now always route through the gate, with an `ensureNonzeroAddress(gate)` guard so a misconfigured/unset gate fails loudly instead of silently no-op'ing a call to `address(0)`.
[a1ffe53](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/a1ffe53a1cfa8b30dc5a39c4d10ac68b04553efd)

contracts/BStock/IBStockLiquidator.sol
Debugger022: I'd lean to keep `vBStock` — it pairs symmetrically with `vDebt` and the bStock market is already unambiguously the collateral.

contracts/BStock/BStockLiquidator.sol
Debugger022: [7df4df2](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/7df4df2dbf8b695e23118e2e16f4befb80f639e3)
`sweep` is `onlyOwner`, so passing a zero address is unlikely in practice — added purely as a safety guard for a clear revert.

contracts/BStock/BStockLiquidator.sol
Debugger022: [7df4df2](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/7df4df2dbf8b695e23118e2e16f4befb80f639e3)

contracts/BStock/BStockLiquidator.sol
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/#discussion_r3503997289 On the naming: agreed, renamed `usdtBefore`/`usdtOut` → `debtBefore`/`debtOut` since the asset is generic `vDebt.underlying()`, not literally USDT.

On forcing output = debt: it's required in flash mode (the flash must be repaid in the borrowed asset), and I'm keeping it in inventory mode too, on purpose. It keeps the `minOut` profitability check in a single unit and replenishes USDT inventory, which is the desired end state for the backstop. Relaxing it in inventory mode wouldn't gain us anything.

Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/#discussion_r3504091921 I'd prefer to keep the seize + swap in one tx. Splitting them reintroduces a price-drift window and leaves the protocol holding the illiquid bStock between steps, the exact thing the atomic flow avoids. Flash mode also can't split (repay is same-tx).

For the "contract might not hold the debt token upfront" case, that's what flash mode is for: it borrows the debt asset and needs no pre-funding, so there's no need for an on-chain pre-swap step.

Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/#discussion_r3504149341 This non-atomic manual flow already exists off-chain: `safe-fallback.ts` generates a Safe batch for exactly this (manual liquidation when the RFQ path is unavailable). So we get the escape hatch without changing the contract.

I recommend keeping `sweep` owner-only. Making it operator-allowed widens the blast radius on an operator-key leak (an operator could pull funds directly), which cuts against the tight permissioning we kept elsewhere.

contracts/BStock/BStockLiquidator.sol
Debugger022: Will set `initialOwner` to the EBrake Emergency Safe (`0xCCa5a587eBDBe80f23c8610F2e53B03158e62948`) directly in initialize.

scripts/bstock/atomic-liquidate.ts
Debugger022: [7d9d1e0](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/7d9d1e04a68bdecb4045c55cc6c75a2d5bbf5231)

contracts/BStock/BStockLiquidator.sol
fred-venus: ok, lets keep router allowlist if the assumption is signer pk can be leaked

scripts/bstock/safe-fallback.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (process.env.TARGET) {
+    target = utils.getAddress(process.env.TARGET);
+  } else if (process.env.ALLOW_PLACEHOLDER === "1") {
+    target = ZERO;
+    console.warn("WARN: TARGET unset — emitting DRAFT with zero address. DO NOT EXECUTE; replace tx #4 `to`/arg.");
+  } else {
+    throw new Error("Missing env TARGET (Binance top-up address). Set it, or ALLOW_PLACEHOLDER=1 for a draft.");
+  }
+
+  // --- build batch ---
```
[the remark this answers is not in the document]
Debugger022: [81fa8df](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/81fa8df81bd995224081c40dadc660aa427a3482)

scripts/bstock/safe-fallback.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  // --- seize math from on-chain truth ---
+  const [seizeErr, seizeTokens]: BigNumber[] = await comptroller.liquidateCalculateSeizeTokens(
+    vDebt.address,
+    vBStock.address,
+    repay,
+  );
+  if (!seizeErr.eq(0)) throw new Error(`liquidateCalculateSeizeTokens error code ${seizeErr}`);
+
+  // Raw bStock from redeeming seizeTokens, FLOORED at the current exchange rate (rate only grows, so
```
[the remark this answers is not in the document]
Debugger022: [81fa8df](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/81fa8df81bd995224081c40dadc660aa427a3482)

contracts/BStock/BStockLiquidator.sol
fred-venus: 👌 let's update the comment accordingly rn still USDT

contracts/BStock/IBStockLiquidator.sol
```diff
@@ -0,0 +1,101 @@
...
+    /// @notice Emitted on a successful liquidation.
+    /// @param borrower The liquidated account.
+    /// @param vBStock The seized bStock collateral market.
+    /// @param repayAmount Debt underlying repaid.
+    /// @param seizedBStock Raw bStock redeemed and sold.
+    /// @param debtOut Debt-asset proceeds of the swap.
+    /// @param flash True if funded by a flash loan, false if from inventory.
+    event Liquidated(
+        address indexed borrower,
+        address indexed vBStock,
```
fred-venus, line 54: also `address vDebtToken` ?

scripts/bstock/lib/native.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/** Earliest order deadline (unix seconds) across a firm quote, or 0 if none. */
+export function quoteDeadline(q: NativeFirmQuote): number {
+  const ds = (q.orders || []).map(o => Number(o.deadlineTimestamp)).filter(n => n > 0);
+  return ds.length ? Math.min(...ds) : 0;
+}
+
+/** Known bStock token addresses on BNB Chain (18 decimals). */
+export const BSTOCK_TOKENS = {
```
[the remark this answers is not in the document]
fred-venus: I suggest we remove this Bstock token check, because in the future we might have even more tokens, dont want to add one more place to update, and also this script is specifically for the liquidation. There will be serious checks before we really execute it, so it is unlikely this will be wrong.

scripts/bstock/native-smoke.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Smoke test for the Native Swap API integration (no chain interaction).
```
fred-venus, line 2: if its purely test, can we rename it a bit so it looks different from the other two scripts

scripts/bstock/atomic-liquidate.ts
```diff
@@ -0,0 +1,220 @@
...
+    // Fork-test path: MOCK_NATIVE = "<router>:<calldata>" pre-encoded against a MockNativeRouter.
+    const [r, data] = process.env.MOCK_NATIVE.split(":");
+    router = ethers.utils.getAddress(r);
+    swapCalldata = data;
+    amountOut = BigNumber.from(process.env.MOCK_OUT || "0");
+  } else {
+    const q = await getFirmQuote({
+      fromAddress: liquidator.address,
+      tokenIn: bStock.address,
+      tokenOut: usdtAddr,
```
fred-venus: we are still specify USDT here not debt token ?

fred-venus: Something just occurred to me:

1. Based on our discussion yesterday, we concluded that the swap output must be a debt token, at least at the smart contract level.
2. However, with native, B-Stock can only be exchanged for USDT (or USDC maybe) 
3. But we can't actually guarantee that the debt token will be USDT, right?

fred-venus: This issue is even more obvious in Flash Mode: 

Since Flash Mode strictly requires the repay token to be equal to the debt token, we run into a problem if the debt token isn't USDT.  After selling the B-Stock, the USDT we receive needs to be converted into the repay token through an additional swap (but for now we dont have that swap ? unless the router is a multi-step router that supports native + pcs/uni

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,257 @@
...
+        //    reverts UNAUTHORIZED, so we route the repay through that Venus Liquidator (it pulls our
+        //    repay and sends us our share of the seized collateral; treasury keeps a cut). The seized
+        //    amount is read as a BALANCE DELTA, so the Liquidator's cut is handled correctly. We guard
+        //    against an unset gate so a misconfig fails loudly instead of silently no-op'ing a call to
+        //    address(0) (a low-level call to a codeless address returns success).
+        uint256 vBefore = params.vBStock.balanceOf(address(this));
+        address gate = comptroller.liquidatorContract();
+        ensureNonzeroAddress(gate);
+        debt.forceApprove(gate, params.repayAmount);
+        ILiquidator(gate).liquidateBorrow(address(params.vDebt), params.borrower, params.repayAmount, params.vBStock);
```
fred-venus: what if the debt token is vbnb market ?

scripts/bstock/safe-fallback.ts
fred-venus, line 1: for this script, also lets consider vbnb market ?

scripts/bstock/lib/native.ts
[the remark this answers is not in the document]
Debugger022: [refactor(bstock): resolve bStock token from orderbook, drop hardcoded…](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/d17b5ba98f1ab8fcc783c494ad3cc7dcf7cdb430)

scripts/bstock/native-smoke.ts
Debugger022: [refactor(bstock): rename native-quote smoke test to native-smoke](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/896dde9217cdac324bb20c83679a47ee41cf0afc)

scripts/bstock/atomic-liquidate.ts
Debugger022: You're right that hop 1 can only ever be `bStock→USDT` (Native quotes bStock against USDT only, confirmed against the live BSC order book). Resolved it with an optional second hop in [1ae2c47](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/1ae2c47d3cd998d8db7cd8d45546cdcf1d7105bd)

- `USDT debt` → single hop: Native bStock→USDT already lands the debt asset.
- `Non-USDT debt` → two hops: hop 1 Native bStock→USDT, hop 2 USDT→debt via a second allowlisted router (AMM/aggregator — KyberSwap/OpenOcean/PancakeV2 in lib/amm.ts). `minOut` is enforced in the final debt asset across the whole chain.

The contract's invariant ("swap output == debt asset") is unchanged — it's just that the "swap" is now one or two hops, and the last hop always produces the debt token. `router2 == address(0)` keeps the exact single-hop behaviour for USDT debt.

This covers flash mode too, which was your sharpest point: the flash must be repaid in the debt asset, so for non-USDT debt, hop 2 converts USDT to the debt token before the callback checks debtOut >= principal + premium and approves the repayment. No multi-step router needed — the two hops are two separate allowlisted router calls, each approval capped at that hop's actual proceeds (seizedBStock for hop 1, the measured intermediate delta for hop 2, so pre-existing inventory is never exposed). Tests cover both inventory and flash with a BTCB debt market.

contracts/BStock/IBStockLiquidator.sol
Debugger022: [17717bc](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/17717bcc08b915afabc920d1c0bea13ecb5bc029)

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,309 @@
...
+
+        // The swap proceeds alone MUST cover principal + premium. Without this, any debt-asset inventory
+        // held by the contract would silently backfill an underwater swap (a real loss), since the
+        // flash repayment is pulled from the total balance, not just the swap output.
+        repayAmounts = new uint256[](1);
+        repayAmounts[0] = amounts[0] + premiums[0];
+        if (debtOut < repayAmounts[0]) revert InsufficientOut(debtOut, repayAmounts[0]);
+
+        // Approve the flashed vToken to pull back principal + premium.
+        IERC20Upgradeable(params.vDebt.underlying()).forceApprove(address(vTokens[0]), repayAmounts[0]);
```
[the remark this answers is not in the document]
Debugger022: @trumpgpt-bot The approval can't be reset before return: the facet pulls the repayment after the callback. executeOperation runs in Phase 2; the transferFrom happens in Phase 3 (_handleFlashLoan → VToken.transferInUnderlyingFlashLoan → doTransferIn), which is after we've returned. Zeroing the allowance here would make that pull fail and revert every flash liquidation. And there's no residual to clean up: _handleFlashLoan caps the pull at borrowedAmount + totalFee, which is exactly the amounts[0] + premiums[0] we approve, so the allowance lands at 0 on its own. Leaving it as-is.

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,309 @@
...
+            IERC20Upgradeable mid = IERC20Upgradeable(params.intermediateToken);
+            uint256 midBefore = mid.balanceOf(address(this));
+            _swap(bStock, params.router, params.swapCalldata, seizedBStock); // hop 1: bStock -> intermediate
+            // Only the hop-1 proceeds are sold onward; any pre-existing intermediate inventory is excluded.
+            uint256 midDelta = mid.balanceOf(address(this)) - midBefore;
+            _swap(mid, params.router2, params.swapCalldata2, midDelta); // hop 2: intermediate -> debt
+        }
+
+        debtOut = debt.balanceOf(address(this)) - debtBefore;
+        if (debtOut < params.minOut) revert InsufficientOut(debtOut, params.minOut);
```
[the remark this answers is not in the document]
Debugger022: [feat(bstock): reject zero minOut in liquidator](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/c9afc4245537db042c039007a449d9b352def3af)

scripts/bstock/atomic-liquidate.ts
```diff
@@ -0,0 +1,281 @@
...
+  // and the fixed-amountIn router pull reverts. 0 today, but governance-settable.
+  let vReceived = seizeTokens;
+  const gate: string = await comptroller.liquidatorContract();
+  if (gate !== ethers.constants.AddressZero) {
+    const venusLiquidator = new Contract(gate, VENUS_LIQUIDATOR_ABI, signer);
+    const liqTreasuryPct: BigNumber = await venusLiquidator.treasuryPercentMantissa();
+    if (!liqTreasuryPct.eq(0)) {
+      const totalIncentive: BigNumber = await comptroller.getEffectiveLiquidationIncentive(borrower, vBStock.address);
+      const bonusAmount = seizeTokens.mul(totalIncentive.sub(ONE)).div(totalIncentive);
+      const ours = bonusAmount.mul(liqTreasuryPct).div(ONE);
```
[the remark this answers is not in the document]
Debugger022: [refactor(bstock): clarify treasury-cut var and hop-2 amountIn constraint](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/087ab6c95fb125daea5503dbfc00a3a504c75e8a)

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,309 @@
...
+                params.intermediateToken == address(debt) ||
+                params.intermediateToken == address(bStock)
+            ) revert InvalidIntermediate();
+
+            IERC20Upgradeable mid = IERC20Upgradeable(params.intermediateToken);
+            uint256 midBefore = mid.balanceOf(address(this));
+            _swap(bStock, params.router, params.swapCalldata, seizedBStock); // hop 1: bStock -> intermediate
+            // Only the hop-1 proceeds are sold onward; any pre-existing intermediate inventory is excluded.
+            uint256 midDelta = mid.balanceOf(address(this)) - midBefore;
+            _swap(mid, params.router2, params.swapCalldata2, midDelta); // hop 2: intermediate -> debt
```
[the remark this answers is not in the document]
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683#discussion_r3511653161

contracts/BStock/BStockLiquidator.sol
Debugger022: [feat(bstock): support native BNB (vBNB) debt](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/d843b27edc6614409774a648f173e9c3151cc19c)

contracts/BStock/BStockLiquidator.sol
```diff
@@ -0,0 +1,426 @@
...
+    ///      `executeOperation` forces the swap proceeds to cover principal + premium — there is no
+    ///      built-in floor tying `debtOut` to `repayAmount`. This asymmetry is intentional: a repay can
+    ///      legitimately out-cost its proceeds (e.g. the Venus Liquidator keeps a treasury cut of the
+    ///      seized collateral, so proceeds land a few % under the repay). `minOut` IS the operator's
+    ///      chosen loss floor for inventory mode — set it to the lowest acceptable debt-asset return.
+    function liquidate(
+        LiquidationParams calldata params
+    ) external override onlyOperator nonReentrant returns (uint256 debtOut) {
+        _validateRouter(params.router);
+        if (params.router2 != address(0)) _validateRouter(params.router2);
```
fred-venus: maybe can just wrap this into `_validateRouter`

scripts/bstock/atomic-liquidate.ts
```diff
@@ -0,0 +1,296 @@
...
+ *   LIQUIDATOR      (req) deployed BStockLiquidator address
+ *   BORROWER        (req) account to liquidate
+ *   VBSTOCK         (req) bStock collateral market (e.g. vTSLAB)
+ *   VDEBT           (req) borrowed market to repay (e.g. vUSDT)
+ *   REPAY_AMOUNT    (req) repay in DEBT underlying, human units
+ *   MODE            "inventory" (default) | "flash"
+ *   SLIPPAGE        Native slippage %, default 0.5
+ *   MIN_OUT_BUFFER  extra haircut on minOut beyond slippage, default 0.5 (%)
+ *   AMM_PROVIDER    hop-2 route source for non-USDT debt: kyberswap (default) | openocean | pcsv2
+ *   WBNB_ADDR       WBNB token, used only for a vBNB (native BNB) debt market (default: BSC WBNB).
```
fred-venus: this doesn't need to be a param ?  its unlikely gonna change

scripts/bstock/safe-fallback.ts
```diff
@@ -0,0 +1,217 @@
...
+  const closeFactor: BigNumber = await comptroller.closeFactorMantissa();
+  console.log(`repay: ${env("REPAY_AMOUNT")} ${debtSym}  (closeFactor=${utils.formatEther(closeFactor)})`);
+
+  // Pool-wide Venus Liquidator gate: when set, a direct vDebt.liquidateBorrow from the Safe reverts
+  // UNAUTHORIZED, so we route the repay through that contract (its permissionless entry). When unset,
+  // the Safe liquidates directly. `repaySpender` is whichever pulls the repay.
+  const gate: string = await comptroller.liquidatorContract();
+  const routed = gate !== ZERO;
+  const repaySpender = routed ? utils.getAddress(gate) : vDebt.address;
+  console.log(
```
fred-venus: lets align the contract behaviour , if `!router` we throw an error

scripts/bstock/atomic-liquidate.ts
```diff
@@ -0,0 +1,296 @@
...
+      swapCalldata2 = data2;
+      intermediateToken = nativeOut;
+    }
+  } else {
+    // Hop 1: Native RFQ always outputs USDT (bStock pairs only with USDT on BSC).
+    const q = await getFirmQuote({
+      fromAddress: liquidator.address,
+      tokenIn: bStock.address,
+      tokenOut: nativeOut,
+      amount: seizedHuman,
```
fred-venus: `seizedHuman` this value actually fluctuates along with the oracle price, so i was thinking should we give some buffer here ? otherwise it could be the actual seized amt is a bit smaller which reverts the whole tx

scripts/bstock/atomic-liquidate.ts
```diff
@@ -0,0 +1,296 @@
...
+  const exchangeRate: BigNumber = await vBStock.exchangeRateStored();
+  const ONE = BigNumber.from(10).pow(18);
+
+  // The seize is routed through the pool-wide Venus Liquidator, which keeps a treasury cut of the
+  // liquidation BONUS (see Liquidator._splitLiquidationIncentive), so this contract receives fewer
+  // vTokens than `seizeTokens`. Deduct that cut, else the precomputed amount overstates our holdings
+  // and the fixed-amountIn router pull reverts. 0 today, but governance-settable.
+  let vReceived = seizeTokens;
+  const gate: string = await comptroller.liquidatorContract();
+  if (gate !== ethers.constants.AddressZero) {
```
fred-venus: same, lets align with contrct, if `!router` just throws an error

scripts/bstock/safe-fallback.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+export async function buildSafeFallbackBatch(provider: providers.Provider) {
+  const safe = utils.getAddress(process.env.SAFE || DEFAULT_SAFE);
+  const borrower = utils.getAddress(env("BORROWER"));
+
+  const vBStock = new Contract(env("VBSTOCK"), VTOKEN_ABI, provider);
+  const vDebt = new Contract(env("VDEBT"), VTOKEN_ABI, provider);
+  const comptroller = new Contract(await vBStock.comptroller(), COMPTROLLER_ABI, provider);
+  const debt = new Contract(await vDebt.underlying(), ERC20_ABI, provider);
```
fred-venus: from ai: vBNB doesnt have `underlying()`

scripts/bstock/safe-fallback.ts
```diff
@@ -0,0 +1,217 @@
...
+    target = ZERO;
+    console.warn("WARN: TARGET unset — emitting DRAFT with zero address. DO NOT EXECUTE; replace tx #4 `to`/arg.");
+  } else {
+    throw new Error("Missing env TARGET (Binance top-up address). Set it, or ALLOW_PLACEHOLDER=1 for a draft.");
+  }
+
+  // --- build batch ---
+  const liquidateTx = routed
+    ? call(gate, "liquidateBorrow(address,address,uint256,address)", [vDebt.address, borrower, repay, vBStock.address])
+    : call(vDebt.address, "liquidateBorrow(address,uint256,address)", [borrower, repay, vBStock.address]);
```
fred-venus: if debt market is is vbnb, should we pass vaule ?

contracts/BStock/BStockLiquidator.sol
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/changes/327a36264eabc90778a1dc3892cf884434cd5cbe

scripts/bstock/atomic-liquidate.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/changes/08f8c328e15be4686888203051d258987aec84e3

scripts/bstock/atomic-liquidate.ts
Debugger022: [fix(bstock):: enforce Venus Liquidator gate checks in atomicLiquidate…](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/28ba5945ef78eac5b3b372279cf27122d39da752)

scripts/bstock/safe-fallback.ts
Debugger022: [fix(bstock):: enforce Venus Liquidator gate checks in atomicLiquidate…](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/28ba5945ef78eac5b3b372279cf27122d39da752)

scripts/bstock/atomic-liquidate.ts
Debugger022: [feat(bstock): add SEIZE_BUFFER parameter to atomicLiquidate for impro…](https://github.com/VenusProtocol/venus-protocol/pull/683/commits/3cef17c4aa485b50f883b6133b3a8c2a34d642d9)

scripts/bstock/safe-fallback.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/commits/812c60004ef2832b93a7eb54a1ab778ec2f874fe

scripts/bstock/safe-fallback.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/683/commits/812c60004ef2832b93a7eb54a1ab778ec2f874fe

contracts/BStock/BStockLiquidator.sol
Debugger022: as discussed with @fred-venus  the owner will be `0x83f426233B358A36953F6951161E76FB7c866a7A`

fred-venus (approved):

fred-venus:
maybe also update the readme (comment) accordingly so that new comes know how to use it

fred-venus (approved):

PR #689, opened 2026-07-06
[VPD-1542] add AccessControlledERC20 token and deployment script

GitGuru7:
## Summary

Introduces a generic, ACM-gated mintable/burnable ERC-20 (`AccessControlledERC20`) and a deployment script for the **Ceffu Custody BTC** receipt token (`vceBTC`). The same non-upgradeable implementation can be reused for different underlying assets — token name, symbol, and decimals are supplied at construction, while minting and burning are gated by the Venus Access Control Manager so only governance-authorized addresses (Timelocks, Guardians) can change supply.

## What Changed

- **`contracts/Tokens/Erc20/AccessControlledERC20.sol`** — new `ERC20` + `Ownable2Step` contract:
  - `mint(address,uint256)` / `burn(address,uint256)` gated via `IAccessControlManagerV8.isAllowedToCall`, reverting with `Unauthorized()`.
  - Configurable `decimals()` set at construction.
  - `setAccessControlManager()` (owner-only) emits `NewAccessControlManager`; zero-address checks on ACM in constructor and setter.
- **`deploy/020-deploy-ceffu-custody-btc.ts`** — new deploy script:
  - Deploys `AccessControlledERC20` as `CeffuCustodyBTC` (`"Ceffu Custody BTC for Venus"` / `vceBTC` / 18 decimals) wired to the deployed `AccessControlManager`.
  - Initiates ownership transfer to `NormalTimelock` (completes via `Ownable2Step` acceptance).
  - Verifies the contract on live networks; skipped on remote networks.


fred-venus:
since its employed by FRV as a collateral token, is it better we keep it there instead of here

contracts/Tokens/VenusERC20/VenusERC20.sol
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: BSD-3-Clause
+pragma solidity 0.8.25;
+
+import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import { Ownable2Step } from "@openzeppelin/contracts/access/Ownable2Step.sol";
+import { IAccessControlManagerV8 } from "@venusprotocol/governance-contracts/contracts/Governance/IAccessControlManagerV8.sol";
+import { ensureNonzeroAddress } from "@venusprotocol/solidity-utilities/contracts/validators.sol";
+
+/**
+ * @title VenusERC20
```
fred-venus, line 10: also the name is too general, context:

1. what we are doing here is trying to create a mirror token of the BTCB token custodied by Ceffu i.e. we mint the same amt of token here to reflect the value stored there
2. it is purely used by FRV, not protocol wise
3. it is going to be solely for similar purposes in the future, maybe another cooperation means another set of FRV contract + ERC20 contract deployment

contracts/Tokens/VenusERC20/VenusERC20.sol
```diff
@@ -0,0 +1,108 @@
...
+ * @notice A generic, non-upgradeable mintable / burnable ERC-20 whose supply is controlled by the Venus
+ * AccessControlManager.
+ * The token name, symbol and decimals are supplied to the constructor so the same implementation can be
+ * reused for different underlying assets. Minting and burning are gated by the AccessControlManager, so only
+ * the addresses that governance grants the `mint(address,uint256)` / `burn(address,uint256)` permissions to
+ * (e.g. the Timelocks and the Guardians) can change the supply.
+ */
+contract VenusERC20 is ERC20, Ownable2Step {
+    /// @notice Number of decimals the token uses, set at construction.
+    uint8 private immutable _decimals;
```
fred-venus, line 21: may i ask whats the purpose of customizing decimals logic here ?  Is the standard erc20 implementation not sufficient or something 🤔?

contracts/Tokens/VenusERC20/VenusERC20.sol
```diff
@@ -0,0 +1,108 @@
...
+ * @title VenusERC20
+ * @author Venus
+ * @notice A generic, non-upgradeable mintable / burnable ERC-20 whose supply is controlled by the Venus
+ * AccessControlManager.
+ * The token name, symbol and decimals are supplied to the constructor so the same implementation can be
+ * reused for different underlying assets. Minting and burning are gated by the AccessControlManager, so only
+ * the addresses that governance grants the `mint(address,uint256)` / `burn(address,uint256)` permissions to
+ * (e.g. the Timelocks and the Guardians) can change the supply.
+ */
+contract VenusERC20 is ERC20, Ownable2Step {
```
fred-venus, line 19: It seems this is purely controlled by the acm. Then what's the point of making it Ownable ?

contracts/Tokens/VenusERC20/VenusERC20.sol
```diff
@@ -0,0 +1,108 @@
...
+     * @notice Destroys `amount_` tokens from `account_`, reducing the total supply.
+     * @param account_ Address from which the tokens are destroyed.
+     * @param amount_ Amount of tokens to be burned.
+     * @custom:access Controlled by AccessControlManager.
+     */
+    function burn(address account_, uint256 amount_) external {
+        _ensureAllowed("burn(address,uint256)");
+        _burn(account_, amount_);
+    }
+
```
fred-venus, line 71: i think it would be helpful if we override `_beforeTokenTransfer` or something to support the ability to freeze or pause, this will help if something severe happens

contracts/Tokens/VenusERC20/VenusERC20.sol
GitGuru7: By default, the OZ ERC-20 implementation hardcodes the decimals to `18`. It would provide us with the flexibility to support tokens with any decimal value.

contracts/Tokens/VenusERC20/VenusERC20.sol
GitGuru7: `onlyOwner` is used for `setAccessControlManager` so that access to updating the ACM address is not gated by the ACM itself. This follows the same approach as the ACM contract,

Well it's very unlikely that we'll ever need to change the ACM contract since it is itself upgradeable, but I think it's better to keep the flexibility to update the address if needed. If you think we should instead make the ACM address immutable, lmk.

contracts/Tokens/VenusERC20/VenusERC20.sol
GitGuru7: initial thought was to keep it reusable so we could use it wherever needed, but it makes sense to move it to the FRV and use it there exclusively. will migrate it there and rename it to **CustodyReceiptToken**

fred-venus:
migrated to FRV repo

PR #694, opened 2026-07-10
[VPD-1589] Add Liquid Mesh as a hop-1 liquidation source

Debugger022:
## Summary

Extends the atomic bStock backstop liquidator with two independent capabilities, without changing the core repay / seize / redeem / `minOut` flow:

- **Liquid Mesh as a second hop-1 RFQ source** alongside Native RFQ. Sources are priced per liquidation and the higher-out one wins; the contract stays source-agnostic.
- **VAI debt support** — for a VAI borrow the hop-2 leg settles through the Peg Stability Module (`swapStableForVAI`), with calldata built locally in `lib/psm.ts`. No aggregator, no contract change.

The contract gains a `routerSpender` mapping so an aggregator whose settlement contract differs from the call target (Liquid Mesh pulls the input via a separate spender) is supported, keeping seize-then-sell atomic.

## Contract

- `routerSpender` mapping + `setRouterSpender` (owner-only, `address(0)` clears). `_swap` approves the router's configured spender, defaulting to the router itself so Native is unchanged. `__gap` 50 -> 49, storage-safe.
- `setRouterSpender` is hardened against misconfiguration: reverts `RouterNotAllowed` unless the router is currently allowlisted, and `SpenderNotContract` for a non-zero spender with no code (the spender receives a live approval during `_swap`, so an EOA is always a misconfig). `setRouter(router, false)` also clears the configured spender, so a stale entry cannot silently reactivate on re-allowlist.
- `RouterSpenderSet` event; source-agnostic natspec.

## Off-chain (`scripts/bstock/`)

**Hop-1 source registry**
- `lib/sources.ts`: a pluggable registry the driver is generic over. `SOURCE=auto` prices every source whose creds are present and takes the higher out; `SOURCE=native,liquidmesh` (or a single name) restricts the set. Adding a source is one adapter and no contract change.
- `lib/liquidmesh.ts`: the Liquid Mesh client (Ed25519 JWT signed over the full path; `/quote`, and `/swap` with `disableSimulate: true` so calldata can be built for a contract that does not yet hold the input). Responses are fully typed and `LM_PRIVATE_KEY_SEED` is validated to decode to exactly 32 bytes with a legible error.
- **Firm vs indicative reconcile**: Native's quote is firm (MM-signed, executes at exactly its `amountOut`), while Liquid Mesh's `/quote` is indicative and the built `/swap` order only guarantees its own floor. Each quote is tagged `firm`, and `build()` exposes the order's `builtFloor`; if an indicative winner's floor undercuts the best firm quote it beat, the driver executes the firm quote instead (logged as `hop-1 reconcile:`).
- **Order-liveness guards**: a Liquid Mesh order with fewer than `LM_MIN_TTL` seconds (default 15) left at build time, or with an implausible millisecond-epoch expiry, is rejected before the settle tx is built rather than reverting on-chain mid-incident. Hop-1 source API calls time out instead of hanging.

**VAI debt (hop-2 via the Peg Stability Module)**
- For a VAI debt the hop-2 leg is the PSM `swapStableForVAI` (USDT -> VAI at the oracle rate), with calldata encoded locally in `lib/psm.ts`. The PSM is just another allowlisted `router2`; the contract is unchanged.
- The builder pre-flights the PSM and aborts legibly (use the Safe fallback) if it `isPaused()`, or if the mint-cap headroom (`vaiMintCap - vaiMinted`) cannot cover the pull — rather than shipping calldata that reverts on-chain. `expectedOut` comes from the real `previewSwapStableForVAI` (USDT prices under $1, so the swap is not 1:1). `MODE=flash` is rejected for VAI (there is no vVAI market to flash from).
- Seize keys off `liquidateVAICalculateSeizeTokens` (VAI is $1, borrower-agnostic core incentive), matching what the gate's VAI branch computes; error messages name the actual seize function called on each branch.
- `safe-fallback.ts` supports VAI as the normal approve -> `liquidateBorrow` batch (the VAI token is resolved via `getVAIAddress()`), detected before the vBNB native-debt fallback so a VAI debt is never misread as native BNB. Since it ships bStock to the CEX with no on-chain swap, it is the manual fallback exactly when the PSM hop is paused or cap-exhausted.

**Seize / proceeds sizing**
- The Venus Liquidator treasury cut is sized off `getEffectiveLiquidationIncentive(borrower, vBStock)` for every debt type, mirroring `Liquidator._splitLiquidationIncentive`; the resolved cut is logged (the live BSC gate keeps 50% of the bonus). For VAI the borrower is always core-pool-locked, so effective equals core there — using effective is gate-parity and future-proofing, and the genuinely divergent case (a non-core ERC20 borrower) is exercised in the fork suite.
- Single-hop `minOut` is derived from the built order's guaranteed floor, not the indicative quote, so an in-slippage Liquid Mesh fill below the quote no longer reverts `InsufficientOut`. This matches the existing two-hop floor sizing.
- Safe-fallback redeem/transfer amounts are haircut by `SEIZE_BUFFER` (default 0.1%) so ordinary oracle price drift between batch generation and signer quorum leaves sweepable dust rather than reverting the redeem.

**Driver / guards**
- `atomic-liquidate.ts` prices all selected sources and settles through the winner.
- `SLIPPAGE` and `MIN_OUT_BUFFER` are validated to `[0,100)` (matching `SEIZE_BUFFER`), so a typo fails loudly before a quote is burned.
- The `getAccountLiquidity` error slot is read and an oracle failure reported distinctly (no longer mislabelled "no shortfall"). The no-shortfall abort is gated behind `ALLOW_NO_SHORTFALL=1`, so a forced liquidation of a healthy account is supported (the contract intentionally does not pre-check liquidatability).
- `verify-lm-fork.ts`: an on-demand fork check that Liquid Mesh `disableSimulate` calldata executes on-chain (dev-time only, never part of an incident).

**Docs**
- README operator runbook: a "which script?" decision flowchart, incident rules of thumb, the source registry and reconcile behaviour, the `setRouterSpender` prerequisite and enforced ordering, and the env knobs (`LM_MIN_TTL`, `SETTLE_TTL_MARGIN`, `ALLOW_NO_SHORTFALL`, `SEIZE_BUFFER`, `PSM_ADDR`).
- The flowchart shows the pre-flight gates (debt-type detection, shortfall / `ALLOW_NO_SHORTFALL`, VAI-forces-inventory), hop-2 routing by debt type (single-hop USDT, AMM for other ERC20, WBNB unwrap for native BNB, PSM for VAI), and the three fallback triggers (no hop-1 source fills, VAI PSM paused/capped, dry-run fails). `native-smoke` is a Native-only probe; Liquid Mesh has no standalone liveness check and is validated inside the atomic dry-run.

## Tests

- `BStockLiquidatorLiquidMesh.ts`: split-spender atomic execution (mocked) + spender-hardening cases (unallowlisted router, EOA spender, spender cleared on de-allowlist).
- `BStockAtomicLiquidate.ts`: source-registry selection (winner, forced source, bad `SOURCE`), firm/indicative reconcile, LM min-TTL and ms-epoch expiry rejection, `SLIPPAGE`/`MIN_OUT_BUFFER` validation, `getAccountLiquidity` error code, `ALLOW_NO_SHORTFALL` forced path, single-hop minOut-from-floor settle, VAI PSM hop-2 (via `MockPSM`), and VAI flash-reject.
- `BStockSafeFallback.ts`: `SEIZE_BUFFER` haircut + range check and the VAI debt batch shape.
- `Fork/BStockLiquidatorFork.ts`: Liquid Mesh split-spender against the real gate + market; two-hop and script-driven VAI liquidations through the real gate and real PSM; paused-PSM refusal; a forced-liquidation scenario under `ALLOW_NO_SHORTFALL`; and effective-incentive divergence on a non-core e-mode ERC20 borrower (proves `getEffectiveLiquidationIncentive != getLiquidationIncentive` on the live diamond and that the driver sizes the cut off the effective incentive).
- Mocks: `MockSpender`, `MockSplitRouter`, `MockPSM`, plus a divergent effective-vs-core incentive and a `getAccountLiquidity` error slot.

## Test plan

- Unit (deterministic): `npx hardhat test tests/hardhat/BStock*.ts` — 106 passing.
- Fork Liquid Mesh: `FORKED_NETWORK=bscmainnet npx hardhat test tests/hardhat/Fork/BStockLiquidatorFork.ts --grep "split router"`
- Fork VAI: `... --grep "VAI"` — two-hop and script-driven VAI settle through the real gate + real PSM; paused-PSM is refused.
- Fork effective-incentive divergence: `... --grep "effective-incentive divergence"`
- Fork forced liquidation: `... --grep "script-driven forced liquidation"` — settles a healthy account under `ALLOW_NO_SHORTFALL`.
- Live Liquid Mesh `disableSimulate` proof (fork at head, chainId 56, LM creds): `npx hardhat run scripts/bstock/verify-lm-fork.ts`.

> Note: the currently-deployed proxy predates both `routerSpender` and the VAI path — a new deploy/upgrade is required before the Liquid Mesh or VAI liquidations work in production.


scripts/bstock/lib/sources.ts
```diff
@@ -0,0 +1,176 @@
...
+
+/** All known hop-1 sources. Add a new aggregator by pushing its adapter here — nothing else changes. */
+export const SOURCES: QuoteSource[] = [nativeSource, liquidMeshSource];
+
+/** Resolve the `SOURCE` env to the active source list: "auto" = every available source, else a subset. */
+export function selectedSources(): QuoteSource[] {
+  const raw = (process.env.SOURCE || "auto").toLowerCase().trim();
+  if (raw === "auto") return SOURCES.filter(s => s.available());
+  const names = raw.split(",").map(s => s.trim());
+  const picked = SOURCES.filter(s => names.includes(s.name));
```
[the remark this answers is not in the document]
Debugger022: [b36c17e](https://github.com/VenusProtocol/venus-protocol/pull/694/commits/b36c17e45b8d08e03085ce8d418a78585a27101e)

scripts/bstock/lib/sources.ts
```diff
@@ -0,0 +1,176 @@
...
+        const minFloor = out.mul(Math.round((100 - a.slippage) * 100)).div(10000);
+        const swap = await lmBuildSwap({
+          userAddress: a.taker,
+          tokenIn: a.tokenIn,
+          tokenOut: a.usdtOut,
+          amountWei: a.weiAmount.toString(),
+          minOutWei: minFloor.toString(),
+          routePlans: quote.routePlans,
+          slippageBps: Math.round(a.slippage * 100),
+        });
```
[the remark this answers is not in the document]
Debugger022: [b36c17e](https://github.com/VenusProtocol/venus-protocol/pull/694/commits/b36c17e45b8d08e03085ce8d418a78585a27101e)

scripts/bstock/atomic-liquidate.ts
```diff
@@ -80,18 +96,87 @@ function env(name: string, required = true): string {
...
+async function pickHop1Source(args: QuoteArgs): Promise<Hop1> {
+  const sources = selectedSources();
+  if (sources.length === 0) {
+    throw new Error(
+      "no hop-1 source available — provide NATIVE_API_KEY and/or LM_API_KEY+LM_PRIVATE_KEY_SEED, or set SOURCE",
+    );
+  }
+
+  // Price every selected source in parallel; a source that errors (API down / no route) drops out.
+  const settled = await Promise.allSettled(sources.map(async s => ({ name: s.name, quote: await s.getQuote(args) })));
```
fred-venus, line 147: should we have a timeout here, otherwise like comment describes if api pending it could block both sources ?

scripts/bstock/atomic-liquidate.ts
```diff
@@ -206,20 +291,26 @@ export async function atomicLiquidate(signer: Signer) {
...
-    deadline = BigNumber.from(quoteDeadline(q)); // settle tx reverts on-chain past the quote's expiry
-    router = q.txRequest.target;
-    swapCalldata = q.txRequest.calldata;
-    const midOut = BigNumber.from(q.amountOut); // USDT out of hop 1
+    const ttl = hop1.deadline.toNumber() - Math.floor(Date.now() / 1000);
+    if (ttl <= 0) throw new Error(`${hop1.source} quote already expired — refetch`);
+    deadline = hop1.deadline; // settle tx reverts on-chain past the quote's expiry
+    router = hop1.router;
+    swapCalldata = hop1.calldata;
+    const midOut = hop1.out; // USDT out of hop 1
```
fred-venus: i understood for that native, output from quote always equal to the amt we received, is it the same case for  liquidMesh ? Unsure if there is a minimumReceived or should we use hop1.out * (100% - slippage)

scripts/bstock/README.md
```diff
@@ -5,24 +5,78 @@ one shared goal: repay a borrower's debt, seize their bStock, and offload it.
...
+- **Don't debug mid-incident.** If the atomic dry-run keeps reverting and the cause isn't obvious in
+  minutes, switch to the Safe fallback instead of burning the liquidation window.
+- `verify-lm-fork.ts` is a **dev-time** fork check (proves LM calldata executes) — never part of an incident.
+
 ---
 
 ## Prerequisites
 
-- **Native API key** (`NATIVE_API_KEY`) for anything that fetches a quote. Never commit it.
+- **Native API key** (`NATIVE_API_KEY`) for anything that fetches a Native quote. Never commit it.
```
fred-venus, line 33: 1. lets confirm is the api key here (both native / lm) ok for long term use, or do we need to reply a new once
2. once done, create a new page here, deploy the contract and update everything there so that the script is ready to use out of box

scripts/bstock/atomic-liquidate.ts
Debugger022: Native's `out` is firm (MM-signed) so received == quoted. LM's `/quote out` is indicative, the binding number is the built order's floor, `builtFloor = out * (1 - slippageBps)` (`sources.ts:120`), so your `out * (1 - slippage)` instinct is right. No underfill is possible either way: the contract reverts if `debtOut < minOut` (`BStockLiquidator.sol:459`), and at default config `minOut` already lands on `0.995·out` == LM's floor. The only caveat is that this holds because `MIN_OUT_BUFFER == SLIPPAGE`; if someone sets the buffer below slippage on a sole-LM run it would false-revert (fails safe, no loss). Can harden `minOut` to derive from `builtFloor` directly if we want to decouple those knobs, but nothing is at risk as-is.

scripts/bstock/README.md
Debugger022: Checked both vendors' docs, neither documents a key expiry or rotation policy. Liquid Mesh issues the key via support, bound to your registered Ed25519 pubkey, and says to store it in a vault (i.e. long-lived); the only short TTL is the per-request JWT (<=2s), which we already mint fresh per call and not the key. 
Native's docs only cover endpoints, nothing on key lifetime. So both read as long-term keys, but since neither publishes a written SLA, we will confirm with each provider before prod.

scripts/bstock/atomic-liquidate.ts
Debugger022: [ae6b50f](https://github.com/VenusProtocol/venus-protocol/pull/694/commits/ae6b50fd5cd33277a16db954ca18ea7b0d6415c9)
Added a shared `fetchWithTimeout` (`AbortSignal.timeout`, env `SOURCE_TIMEOUT_MS`, default 8s) to both clients, so a hung source aborts and drops out of the race instead of blocking the live one. Covered by `BStockSourceTimeout.ts`.

scripts/bstock/README.md
fred-venus: cool then, i forgot to attach the doc

1. lets create a page here under Credentials section 
    - https://app.notion.com/p/Venus-Team-Controlled-Multisig-EOA-Credentials-319b4e5771bb8002a7c8cfa3df3bd93b
2. as for the format can ref other doc

scripts/bstock/atomic-liquidate.ts
fred-venus: lets just keep it then.

In fact, my original thought is that the actual amount of tokens we received from hop1 could be slightly smaller than this value. This could cause the hop2 swap to fail because the amount we used in the call request is more than we can actually provide. But yeah, like u said, no harm, maybe add a comment here indicating if this fail can slightly lower this value or use builtFloor value ?

Debugger022: [591f6c7](https://github.com/VenusProtocol/venus-protocol/pull/694/commits/591f6c73c02f548de8d8e384ad83131b4abdf20a) went with `builtFloor`.

scripts/bstock/atomic-liquidate.ts
```diff
@@ -223,28 +328,46 @@ export async function atomicLiquidate(signer: Signer) {
...
 
     if (twoHop) {
       // Hop 2: convert the hop-1 USDT to the (non-USDT) debt asset via an allowlisted AMM/aggregator.
+      // Size this leg off the hop-1 FLOOR, not the indicative `out`: on-chain the contract approves router2
+      // for the ACTUAL hop-1 delta (`midDelta`), while this calldata bakes in a fixed `amountIn`. Quote it
+      // at `out` and an indicative source that fills even slightly under would leave the router pulling
+      // more than the approval — hop 2 reverts on allowance. `floor <= midDelta` always holds, so the pull
+      // always fits. Any surplus (`midDelta - floor`, bounded by slippage) stays as USDT inventory and the
+      // contract emits `PartialSwapLeftover` for it — sweepable, not lost.
       const amm = await getAmmSwap(
```
fred-venus: for VAI, its gonna be `swapStableForVAI` ? But where is the code to construct the calldata ?

Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/694/changes/3ab7e4f062f789e1f3aba98373ff901e80cae4f0 https://github.com/VenusProtocol/venus-protocol/pull/694/changes/1baf78f22f77016ade94a0ae143368b175c7a5b1 https://github.com/VenusProtocol/venus-protocol/pull/694/changes/111af24985ba3d34fadddf9332b9baa9dd7d598f

done.

Debugger022:
@claude please review

PR #707, opened 2026-08-20
Feat/bstock spoke liquidation

Debugger022:
## What this does

`BStockLiquidator` is the atomic backstop liquidator for bStock collateral. Until now it only worked against the Core pool: it read the pool-wide gate off the Core Comptroller and flash borrowed from the Core diamond. A hub-funded spoke pool (a `SpokeComptroller` with a USDT debt leg and bStock collateral) has neither of those.

This makes the same contract serve both pools. No second deployment, no second address to track.

## How the pool is chosen

`_resolvePool` reads the **collateral market's own** comptroller, then:

- equals the Core immutable → Core path, unchanged, byte for byte
- otherwise → must be owner-allowlisted, must answer `isComptroller()`, and the pool must list **both** legs, or it reverts (`ComptrollerNotAllowed` / `MarketNotInPool`)

The debt leg's comptroller is deliberately never read. Anchoring on one leg and asking the pool about the other is what stops a forged `comptroller()` from routing a hostile market.

On the spoke path the repay goes straight to the debt market (no gate), and flash principal is borrowed from a Core market keyed by the debt token (`coreFlashSource`), since the spoke has no flash lender of its own.

## Storage

Two appended slots, verified against the deployed implementation:

| Slot | Field |
| --- | --- |
| 204 | `isAllowedComptroller` |
| 205 | `coreFlashSource` |
| 206 | `__gap`, `uint256[49]` → `uint256[47]` |

Last slot stays 252. Both mappings start empty, so an upgraded proxy behaves exactly as before until a pool is deliberately enabled. Runtime size 13,705 bytes.

## Testing

233 tests, all green.

| Layer | Tests | Covers |
| --- | --- | --- |
| Unit, Core | 52 | pre-existing behaviour, unchanged |
| Unit, spoke | 51 | isolated branch + mock fidelity vs the real artifacts |
| Unit, layout | 6 | append-only storage, gap arithmetic, EIP-170 |
| Unit, scripts | 64 | operator tooling |
| Fork, Core | 26 | live proxy upgraded in place, then every Core scenario re-run |
| Fork, spoke | 24 | real pool contracts deployed into a bscmainnet fork |
| Fork, matrix | 10 | unchanged |

The Core fork suite upgrades the **live bscmainnet proxy** through its own admin and re-runs every Core scenario against it, so the whole file doubles as a post-upgrade regression.

The spoke fork suite deploys the real `SpokeComptroller` and real isolated `VToken`s into the fork (vendored artifacts) on top of the live AccessControlManager, timelock and oracle. Only `poolRegistry` and `protocolShareReserve` are stand-ins, neither on the liquidation path. Seize amounts are asserted against the pool's own `liquidateCalculateSeizeTokens` using the exact double-floor the market uses, not an approximation.


contracts/BStock/BStockLiquidator.sol
```diff
@@ -202,6 +239,32 @@ contract BStockLiquidator is
         emit RouterSpenderSet(router, spender);
     }
 
+    /// @inheritdoc IBStockLiquidator
+    function setAllowedComptroller(address comptroller_, bool allowed) external override onlyOwner {
```
GitGuru7: How about we use a pool registry-based check instead of maintaining a separate allowlist?

henryleunghc:
## Merge Status: ✅ MERGEABLE

PR is code-complete and mergeable. Has review feedback from claude and GitGuru7. Address comments and ready to merge.

contracts/BStock/BStockLiquidator.sol
Debugger022: Yeah, that would be fine. For now, we only have one pool registry, so a single `poolRegistry` address is enough. But if we add another non-Core pool family later, such as the existing isolated-pools registry, we would need something like `mapping(address => bool) isAllowedPoolRegistry` to support multiple registries.

Debugger022: [refactor(bstock): resolve spoke pools through a PoolRegistry](https://github.com/VenusProtocol/venus-protocol/pull/707/commits/cb3311386ec41e44c2fa24cdf9815f341df2cac1)

GitGuru7: What I meant was that we could skip the allowlist entirely and just check whether the Comptroller is registered in our Pool Registry. If it is, then we know it belongs to us, and any calls made through this Liquidator are anyway gated to us.

I don’t think we would deploy multiple Pool Registries, at least not on the same chain. Multiple pools are isolated based on their different Comptroller deployments, and these are then registered in a single registry.

scripts/bstock/atomic-liquidate.ts
```diff
@@ -234,20 +358,28 @@ export async function atomicLiquidate(signer: Signer) {
...
+    vaiControllerAddr = await comptroller.vaiController();
+    isVai = vDebt.address.toLowerCase() === vaiControllerAddr.toLowerCase();
+    if (isVai) {
+      debtAddr = await new Contract(vaiControllerAddr, VAI_CONTROLLER_ABI, signer).getVAIAddress();
+    } else {
+      try {
+        debtAddr = await vDebt.underlying();
+      } catch {
+        isBnb = true;
+        debtAddr = ethers.utils.getAddress(process.env.WBNB_ADDR || BSC_WBNB);
```
GitGuru7: I would prefer having the actual vBNB address and comparing against it so we know it’s BNB, we can have a constant defined with the vBNB address, and this can simply be handled with an `else if` for the BNB check.

scripts/bstock/atomic-liquidate.ts
```diff
@@ -291,72 +536,117 @@ export async function atomicLiquidate(signer: Signer) {
...
+  //             `VToken._liquidateBorrowFresh` calls. There is no borrower argument because there are no
+  //             per-borrower pools here; the incentive is keyed on the COLLATERAL market. Calling Core's
+  //             4-arg shape against an isolated comptroller reverts — it has no such function and no fallback.
   const seizeFn = isVai ? "liquidateVAICalculateSeizeTokens" : "liquidateCalculateSeizeTokens";
   const [seizeErr, seizeTokens]: BigNumber[] = isVai
     ? await comptroller.liquidateVAICalculateSeizeTokens(vBStock.address, repay)
-    : await comptroller.liquidateCalculateSeizeTokens(borrower, vDebt.address, vBStock.address, repay);
+    : isCore
+      ? await comptroller.liquidateCalculateSeizeTokens(borrower, vDebt.address, vBStock.address, repay)
+      : await comptroller.liquidateCalculateSeizeTokens(vDebt.address, vBStock.address, repay);
```
GitGuru7: then this could also revert on IL code base similar to getAccountLiquidity, so should we add try catch here as well ?

scripts/bstock/atomic-liquidate.ts
GitGuru7, line 653: USDT address would remain same so maybe can skip taking it from the ENV entirely, ? this only adds the scope for passing wrong address.

scripts/bstock/atomic-liquidate.ts
```diff
@@ -262,18 +394,131 @@ export async function atomicLiquidate(signer: Signer) {
...
+    // which is what keeps its own redeem off the liquidity check and off the deviation-bounded oracle.)
+    if (!(await comptroller.checkMembership(borrower, vBStock.address))) {
+      throw new Error(
+        `${borrower} has not entered ${vBStock.address} as collateral — preSeizeHook reverts MarketNotCollateral`,
+      );
+    }
+
+    const borrowBalance: BigNumber = await vDebt.borrowBalanceStored(borrower);
+    const closeFactor: BigNumber = await comptroller.closeFactorMantissa();
+    const maxRepay = forcedIsolated ? borrowBalance : borrowBalance.mul(closeFactor).div(ONE18);
```
GitGuru7: then should also pre check this closeFactor based maxRepay for Core as well ?

scripts/bstock/safe-fallback.ts
```diff
@@ -108,29 +126,68 @@ export async function buildSafeFallbackBatch(provider: providers.Provider) {
...
+  // liquidateBorrow signature and whether a treasury cut applies. Probed directly rather than compared
+  // against a hardcoded Core address, because this script never touches the BStockLiquidator and so has no
+  // Core immutable to compare against. Core answers; an isolated comptroller has no such function and no
+  // fallback, so the call fails — and that failure IS the signal.
+  let isCore = true;
+  let gateProbe = ZERO;
+  try {
+    gateProbe = await new Contract(poolAddr, GATE_PROBE_ABI, provider).liquidatorContract();
+  } catch {
+    isCore = false;
```
GitGuru7: better compare with constant CORE_COMPTROLLER ADD

scripts/bstock/safe-fallback.ts
```diff
@@ -94,6 +94,24 @@ const COMPTROLLER_ABI = [
 const VAI_CONTROLLER_ABI = ["function getVAIAddress() view returns (address)"];
 const LIQUIDATOR_ABI = ["function treasuryPercentMantissa() view returns (uint256)"];
 
+// Isolated-pools `Comptroller` / `SpokeComptroller`. Kept DISJOINT from COMPTROLLER_ABI: none of Core's gate
+// reads exist here, and the pool has no fallback to answer them with a plausible zero.
+// `liquidationIncentiveMantissa()` is deliberately absent — it answers for msg.sender, and an eth_call
+// carries no from-address, so it would quietly return the pool-wide default and hide a per-market override.
+const ISOLATED_COMPTROLLER_ABI = [
```
GitGuru7: better to move all the abis to a separate file and then share among this fallback and atomic liquidator

contracts/BStock/BStockLiquidator.sol
Debugger022: correct changes are done in above commit

scripts/bstock/atomic-liquidate.ts
Debugger022: This env isn't for mainnet, it's what the tests rely on. The non-fork tests deploy a mock USDT and pass its address in through USDT_ADDR ([here](https://github.com/VenusProtocol/venus-protocol/blob/cb3311386ec41e44c2fa24cdf9815f341df2cac1/tests/hardhat/BStockAtomicLiquidate.ts#L77)), same with WBNB_ADDR for mock WBNB. If we hardcode the constant, the script always looks for the real BSC USDT, which doesn't exist on the test chain, so those tests fail.

On the wrong-address risk: when the env isn't set it already falls back to BSC_USDT, so you'd have to set it on purpose to get a wrong address. Same reasoning is [written inline](https://github.com/VenusProtocol/venus-protocol/blob/cb3311386ec41e44c2fa24cdf9815f341df2cac1/scripts/bstock/atomic-liquidate.ts#L356-L357).

scripts/bstock/atomic-liquidate.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/707/commits/c8fd2aac3cf31b3afa2f506aa6f62e7070f91311

scripts/bstock/atomic-liquidate.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/707/commits/c8fd2aac3cf31b3afa2f506aa6f62e7070f91311

scripts/bstock/atomic-liquidate.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/707/commits/c8fd2aac3cf31b3afa2f506aa6f62e7070f91311

scripts/bstock/safe-fallback.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/707/commits/c8fd2aac3cf31b3afa2f506aa6f62e7070f91311

scripts/bstock/safe-fallback.ts
Debugger022: https://github.com/VenusProtocol/venus-protocol/pull/707/commits/fe2b11cfd3f3bc4f47320d43f456f006dc50b55e

contracts/BStock/BStockLiquidator.sol
```diff
@@ -202,6 +254,33 @@ contract BStockLiquidator is
...
+            // Probe the getter every isolated liquidation depends on, so a wrong address fails here rather
+            // than mid-liquidation. Only that the call answers and decodes matters, not what it returns.
+            IPoolRegistry(poolRegistry_).getPoolByComptroller(address(0));
+        }
+        emit PoolRegistrySet(poolRegistry, poolRegistry_);
+        poolRegistry = poolRegistry_;
+    }
+
+    /// @inheritdoc IBStockLiquidator
+    function setCoreFlashSource(address debtToken, IVBep20 vToken) external override onlyOwner {
```
fred-venus, line 271: not sure if we have getter to reduce vToken addr from core pool controller so that we could avoid remove the additional operational work here 🤔

fred-venus (approved):

fred-venus:
We probably wanted to update the name as well, its now not limited to BStock

fred-venus:
On a second thought, nvm, lets keep BStockliquidator, we could rename when there is really some new asset being added into spoke pool and that asset needs us to take care

fred-venus:
ok, 3 more from ai:


1. safe-fallback.ts: isolated pre-flight parity

atomic-liquidate.ts pre-checks minLiquidatableCollateral, the LIQUIDATE/SEIZE/REDEEM pauses and checkMembership for an isolated pool; safe-fallback.ts checks none of them. All are deterministic reverts, so by this script's own TooMuchRepay reasoning they should hard-abort here too, or the batch fails after a full signing round. Suggest a shared isolated pre-flight helper in lib/ used by both scripts.

---

2. forced vs ALLOW_NO_SHORTFALL

Now that both scripts read the forced flag, ALLOW_NO_SHORTFALL is redundant when the flag is on and harmful when it is off (the tx reverts InsufficientShortfall after a firm quote is spent). Suggest dropping the env var: shortfall == 0 and not forced → hard abort. The flag is already governance-signed, so the extra confirmation adds nothing.

---

3. README still says the Safe script picks Core vs isolated by probing liquidatorContract():

The script decides this by probing the pool for a `liquidatorContract()` gate rather
than by address, so it needs no extra env.

The code does the opposite — an address comparison against CORE_COMPTROLLER — and even has a test that a pool answering liquidatorContract() must not be treated as Core. The fork comment at BStockLiquidatorSpokeFork.ts:676 and the safe-fallback.ts file header still describe the old probe-and-revert path too.

contracts/BStock/BStockLiquidator.sol
Debugger022: no such getter exists on Core.
- Core's Comptroller only indexes vToken → market data (`markets(vToken)`, `getAllMarkets()`), no underlying → vToken lookup. `getVTokenForAsset` exists on the isolated-pools PoolRegistry, but Core isn't registered in a PoolRegistry so it can't answer for Core markets.
- The only on-chain substitute is scanning `getAllMarkets()` and calling `underlying()` on each. That's 55 markets today, and the vBNB entry returns no data so the call reverts on decode.

Debugger022:
> ok, 3 more from ai:
> 
> 1. safe-fallback.ts: isolated pre-flight parity
> 
> atomic-liquidate.ts pre-checks minLiquidatableCollateral, the LIQUIDATE/SEIZE/REDEEM pauses and checkMembership for an isolated pool; safe-fallback.ts checks none of them. All are deterministic reverts, so by this script's own TooMuchRepay reasoning they should hard-abort here too, or the batch fails after a full signing round. Suggest a shared isolated pre-flight helper in lib/ used by both scripts.
> 
> 2. forced vs ALLOW_NO_SHORTFALL
> 
> Now that both scripts read the forced flag, ALLOW_NO_SHORTFALL is redundant when the flag is on and harmful when it is off (the tx reverts InsufficientShortfall after a firm quote is spent). Suggest dropping the env var: shortfall == 0 and not forced → hard abort. The flag is already governance-signed, so the extra confirmation adds nothing.
> 
> 3. README still says the Safe script picks Core vs isolated by probing liquidatorContract():
> 
> The script decides this by probing the pool for a `liquidatorContract()` gate rather than by address, so it needs no extra env.
> 
> The code does the opposite — an address comparison against CORE_COMPTROLLER — and even has a test that a pool answering liquidatorContract() must not be treated as Core. The fork comment at BStockLiquidatorSpokeFork.ts:676 and the safe-fallback.ts file header still describe the old probe-and-revert path too.

https://github.com/VenusProtocol/venus-protocol/pull/707/commits/664f9361f497328c6bdb6b6f20db1daf85929072

1. Pre-flight parity: partly valid, fixed pool-agnostic since Core has the same actionPaused/checkMembership gates; minLiquidatableCollateral is oracle-priced so it warns in the fallback rather than aborting.
2. ALLOW_NO_SHORTFALL: not valid, not dropped: it is the only way to serve a borderline account whose shortfall the stored-balance read understates; misleading message reworded.
3. README: stale, correct, fixed along with the fork test comment.

fred-venus (approved):

PR #708, opened 2026-08-24
feat(markets): add vhUSDT, vhUSDC and vhU market deployment

Debugger022:
- Registers the three Venus Hub receipt tokens (24-decimal ERC4626) so 011-deploy-markets can resolve them by symbol
- Interest rate model params resolve to the plain JumpRateModel_base0bps_slope900bps_jump20000bps_kink5000bps_bpy70080000 at 0x6463ab80, which the listing VIP also sets explicitly
- riskParameters and initialSupply are recorded for reference only; no deploy script reads them and the listing VIP sets the live values


fred-venus:
lgtm, just one from ai:

Mainnet aggregates not regenerated (148 and 153, independently) — bscmainnet.json / bscmainnet_addresses.json have no vvhUSDT/vvhUSDC/vvhU while the testnet half of the same diff was regenerated. Same missed export as oracle#320.

fred-venus (approved):

Debugger022:
> lgtm, just one from ai:
> 
> Mainnet aggregates not regenerated (148 and 153, independently) — bscmainnet.json / bscmainnet_addresses.json have no vvhUSDT/vvhUSDC/vvhU while the testnet half of the same diff was regenerated. Same missed export as oracle#320.

it's because of github glitch, now these files exits 
https://github.com/VenusProtocol/venus-protocol/pull/708/changes/9b11add2ccb983a5000f06827bf86508115c2eed
image

PR #709, opened 2026-09-03
refactor(prime-update-scores): target PrimeV2, split into index/update stages

fred-venus:
## Why

The next Prime allocation VIP (vips PR #763) calls `PrimeV2.addMarket(vU, 0, 2e18)`. On PrimeV2 that sets `pendingScoreUpdates = totalTokens` (currently 500) and, until the round is drained with `updateScores`, `claimPrime/claimPrimeBatch/issue/issueBatch/burn/burnBatch` revert with `ScoreUpdateInProgress`, no holder has a vU score (U rewards land in `undistributedReward`), and the monthly cycle cannot run. VIP-637's two `addMarket` calls ran at 0 holders, so this is the first real round on PrimeV2.

`scripts/prime-update-scores.ts` is the only tool for this, but it targets the legacy `Prime` deployment (name, ABI, deploy block 33,264,762) and does scan + send in one go.

## What

- Target `PrimeV2`; scan Mint/Burn from the proxy deploy block in the hardhat-deploy artifact (bscmainnet 107,040,035 / bsctestnet 112,385,727). Uses the deployment ABI, so the `Mint(address)` vs legacy `Mint(address,bool)` difference doesn't matter — both name the arg `user`.
- Two explicit stages, so the holder list is reviewed before any transaction:
  - `STAGE=index` — replay events into the current holder set, check it equals `totalTokens()`, write `HOLDERS_FILE` with the current `nextScoreUpdateRoundId`, stop. Mismatch is an error.
  - `STAGE=update` — refuses to run without `HOLDERS_FILE` or if the round id differs from the file's; skips users already done this round via `isScoreUpdated`; sends in batches of `BATCH_SIZE` (default 14, BSC's 16.78M gas cap ÷ ~0.93M per user).
- The list cannot go stale during a round: every mint/burn entry point (`claimPrime`, `claimPrimeBatch`, `issue`, `issueBatch`, `burn`, `burnBatch`) reverts `ScoreUpdateInProgress` while `pendingScoreUpdates > 0`. The only stale window is index → round open, and the round-id check closes it.
- Dropped the two retry loops: the run is idempotent through `isScoreUpdated`, so a failed run is simply re-run. 203 lines → 85.

## Verified

- `yarn lint` clean; `hardhat test tests/hardhat/Scripts/*` — 46 passing (unit + e2e, unchanged).
- `STAGE=index` on bscmainnet: 754 events → 500 holders == `totalTokens()`, all 500 `isPrimeHolder == true`, the 13 minted / 13 burned in the Sep 1 cycle are in / out. Matches an independently produced list address-for-address. Re-run 2026-09-04 after the review fixes: same 500, round 2; `BLOCK_RANGE=50000` on NodeReal takes ~50 s.
- `STAGE=update DRY_RUN=1` on bscmainnet: reads the file, reports `pendingScoreUpdates = 0 — nothing to do` (no round open yet).

## Runbook after the VIP executes

```
export ARCHIVE_NODE_bscmainnet=… DEPLOYER_PRIVATE_KEY=…
STAGE=index  npx hardhat run scripts/prime-update-scores.ts --network bscmainnet   # after the VIP executes (round open)
STAGE=update npx hardhat run scripts/prime-update-scores.ts --network bscmainnet   # ~36 txs
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/prime-update-scores.ts
+++ b/scripts/prime-update-scores.ts
@@ -1,25 +1,14 @@
 /**
- * Drains PrimeV2's pendingScoreUpdates queue after addMarket / updateAlpha /
- * updateMultipliers. Two stages, so the holder list is reviewed before any tx:
+ * Drains PrimeV2's pendingScoreUpdates after addMarket / updateAlpha / updateMultipliers.
  *
- *   STAGE=index  npx hardhat run scripts/prime-update-scores.ts --network bscmainnet
- *     Replays Mint/Burn events since the PrimeV2 deploy block into the current
- *     holder set, checks the count against totalTokens(), writes HOLDERS_FILE.
+ *   STAGE=index   replay Mint/Burn into the holder set, check it against totalTokens(), write HOLDERS_FILE
+ *   STAGE=update  read HOLDERS_FILE, send updateScores in batches; skips users already done, so re-runnable
  *
- *   STAGE=update npx hardhat run scripts/prime-update-scores.ts --network bscmainnet
- *     Reads HOLDERS_FILE, skips users already done this round (isScoreUpdated),
- *     sends updateScores in batches. Re-runnable: a crashed run just resumes.
- *
- * WARNING — the holder list must match the contract at the moment STAGE=update runs.
- * The monthly Prime cycle (keeper burnBatch + issueBatch, 1st of the month ~00:00 UTC)
- * swaps holders, and a permissionless claimPrime can add one at any time. With a stale
- * list, updateScores never sees the new holders, pendingScoreUpdates never reaches 0, and
- * claimPrime/issue/burn keep reverting. Equal-count churn (13 out, 13 in) is invisible to
- * the totalTokens() check below. So: never run STAGE=update across the monthly cycle, and
- * re-run STAGE=index right before STAGE=update whenever time has passed since indexing.
- *
- * Configuration (env var, or constant below) — one line each: meaning · stage · default
+ * The list cannot go stale during a round: every mint/burn entry point reverts ScoreUpdateInProgress
+ * while pendingScoreUpdates > 0. index records the round id and update refuses any other round.
+ * Drain promptly: an open round also blocks the monthly keeper burnBatch/issueBatch.
  */
+import type { Event } from "ethers";
 import * as fs from "fs";
 import { deployments, ethers } from "hardhat";
 
@@ -28,11 +17,10 @@ const STAGE = process.env.STAGE;
 /** hardhat-deploy name the Prime address + ABI are read from (deployments/<network>/PrimeV2.json). Both stages. */
 const PRIME_DEPLOYMENT = "PrimeV2";
 /** Where the holder list is written (index) and read from (update). Default .prime-holders-<chainId>.json in cwd. */
-const HOLDERS_FILE = process.env.HOLDERS_FILE; // resolved in main() once chainId is known
-/** First block of the Mint/Burn scan. index only. Default: PrimeV2 proxy creation on bscmainnet
- *  (tx 0x41d505b974f7435664281ce32a39df8bcfee74e5a36548c2710f0deede74b0fe) — no Prime events exist before it; 0 elsewhere. */
+const HOLDERS_FILE = process.env.HOLDERS_FILE;
+/** First block of the Mint/Burn scan. index only. Default: the proxy's deployment block from the
+ *  hardhat-deploy artifact (bscmainnet 107,040,035 / bsctestnet 112,385,727) — no events exist before it. */
 const FROM_BLOCK = process.env.FROM_BLOCK ? Number(process.env.FROM_BLOCK) : undefined;
-const PRIMEV2_DEPLOY_BLOCK_BSCMAINNET = 107_040_035;
 /** Blocks per eth_getLogs call. index only. Default 5000 (safe on public RPCs; NodeReal takes 50000). */
 const BLOCK_RANGE = Number(process.env.BLOCK_RANGE ?? 5000);
 /** Users per updateScores tx. update only. Default 14: BSC caps a tx at 16,777,216 gas and one user
@@ -63,6 +51,10 @@ export function eventsToHolders(events: HolderEvent[]): string[] {
 }
 
 /** Both Prime and PrimeV2 name the event arg `user`, so this works for either ABI. */
+const toEvent =
+  (kind: HolderEvent["kind"]) =>
+  (e: Event): HolderEvent => ({ kind, user: e.args?.user, blockNumber: e.blockNumber, logIndex: e.logIndex });
+
 export async function fetchHolderEvents(
   prime: any,
   fromBlock: number,
@@ -74,11 +66,11 @@ export async function fetchHolderEvents(
   const events: HolderEvent[] = [];
   for (let from = fromBlock; from <= toBlock; from += blockRange) {
     const to = Math.min(from + blockRange - 1, toBlock);
-    for (const kind of ["Mint", "Burn"] as const) {
-      for (const e of await prime.queryFilter(prime.filters[kind](), from, to)) {
-        events.push({ kind, user: e.args.user, blockNumber: e.blockNumber, logIndex: e.logIndex });
-      }
-    }
+    const [mints, burns] = await Promise.all([
+      prime.queryFilter(prime.filters.Mint(), from, to),
+      prime.queryFilter(prime.filters.Burn(), from, to),
+    ]);
+    events.push(...mints.map(toEvent("Mint")), ...burns.map(toEvent("Burn")));
     log(`  scanned ${from}-${to}: ${events.length} events so far`);
   }
   return events;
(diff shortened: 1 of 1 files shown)


scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,154 @@
...
-  await runUpdate(prime, holders, { dryRun, log: m => console.log(m) });
+  const saved = JSON.parse(fs.readFileSync(file, "utf8"));
+  if (saved.address.toLowerCase() !== dep.address.toLowerCase() || saved.chainId !== chainId) {
+    throw new Error(`${file} was indexed for ${saved.address} on chain ${saved.chainId}`);
+  }
+  const onChain = Number(await prime.totalTokens());
+  console.log(`${saved.holders.length} holders indexed at block ${saved.toBlock}; on-chain totalTokens now ${onChain}`);
+  if (onChain !== saved.holders.length)
+    console.warn("WARNING: holder set changed since indexing — burned users are skipped, new mints are missing.");
+  await runUpdate(prime, saved.holders, {
```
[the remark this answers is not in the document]
fred-venus: Addressed in b589b3c9 as documentation rather than logic: the header now carries a WARNING that the list must match the contract at run time, that the monthly burnBatch/issueBatch cycle (1st of month ~00:00 UTC) and any claimPrime make it stale, that equal-count churn is invisible to the totalTokens() check, and that the operator must not run STAGE=update across the cycle and should re-run STAGE=index right before STAGE=update. A stricter check (re-scan events since `toBlock` inside STAGE=update) is straightforward if we want it enforced in code.

scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,171 @@
...
+  const file = HOLDERS_FILE ?? `.prime-holders-${chainId}.json`;
+  console.log(`${PRIME_DEPLOYMENT} ${dep.address} chainId ${chainId}, holders file ${file}`);
+
+  if (STAGE === "index") {
+    const fromBlock = FROM_BLOCK ?? (chainId === 56 ? PRIMEV2_DEPLOY_BLOCK_BSCMAINNET : 0);
+    const toBlock = await ethers.provider.getBlockNumber();
+    const events = await fetchHolderEvents(prime, fromBlock, BLOCK_RANGE, toBlock, console.log);
+    const holders = eventsToHolders(events);
+    const onChain = Number(await prime.totalTokens());
+    fs.writeFileSync(file, JSON.stringify({ address: dep.address, chainId, toBlock, holders }, null, 2));
```
Debugger022: may be we can check first then write it. 

```
if (holders.length !== onChain)
  throw new Error("indexed holder count != totalTokens — check FROM_BLOCK / RPC gaps");
fs.writeFileSync(file, JSON.stringify({ address: dep.address, chainId, toBlock, holders }, null, 2));
```

scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,171 @@
...
- * Prime exposes no on-chain holder enumeration, so this script reconstructs
- * the current holder set from `Mint` and `Burn` events, then calls
- * `updateScores` in batches sized to `maxLoopsLimit()` until the queue drains.
+ *   STAGE=update npx hardhat run scripts/prime-update-scores.ts --network bscmainnet
+ *     Reads HOLDERS_FILE, skips users already done this round (isScoreUpdated),
+ *     sends updateScores in batches. Re-runnable: a crashed run just resumes.
  *
- * Resumes correctly across runs: per-user, per-round dedup via
- * `isScoreUpdated[roundId][user]` is read on-chain.
+ * WARNING — the holder list must match the contract at the moment STAGE=update runs.
```
Debugger022: While a round is open the holder set is frozen: `claimPrime`, `claimPrimeBatch`, `issue`, `issueBatch`, `burn`, `burnBatch` all revert with `ScoreUpdateInProgress` when `pendingScoreUpdates > 0`, and those are the only callers of `_mint`/`_burn`.

So the stale window is index to round-open, not during the update. Index after the round is open and the list cannot go stale, which may remove the need for the re-scan.

Should the header say that? It currently says the cycle and `claimPrime` can change holders during the update. Also worth noting: an undrained round blocks the monthly `burnBatch`/`issueBatch`.

scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,171 @@
...
-    // Mint(address indexed user, bool isIrrevocable): we only need `user`.
-    for (const m of mints) {
-      events.push({ kind: "Mint", user: m.args!.user, blockNumber: m.blockNumber, logIndex: m.logIndex });
-    }
-    // Burn(address indexed user)
-    for (const b of burns) {
-      events.push({ kind: "Burn", user: b.args!.user, blockNumber: b.blockNumber, logIndex: b.logIndex });
+  for (let from = fromBlock; from <= toBlock; from += blockRange) {
+    const to = Math.min(from + blockRange - 1, toBlock);
+    for (const kind of ["Mint", "Burn"] as const) {
```
Debugger022: These were a `Promise.all` before. Serial is about 620ms per range against about 350ms, so roughly 26 min instead of 15 for a full mainnet index. Any reason to keep them sequential? We already run 20 concurrent reads at `READ_CONCURRENCY`.

scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,171 @@
...
-    for (let j = 0; j < slice.length; j++) {
-      if (!flags[j]) remaining.push(slice[j]);
-    }
+    return { pendingBefore, pendingAfter: 0n, batchesSent: 0, usersUpdated: 0 };
   }
+  const round = await prime.nextScoreUpdateRoundId();
+  const maxLoops = Number(await prime.maxLoopsLimit());
+  log(`pendingScoreUpdates ${pendingBefore}, round ${round}, maxLoopsLimit ${maxLoops}`);
+
+  // ponytail: READ_CONCURRENCY reads in flight; fine for 500 holders, use a multicall if it ever isn't.
```
Debugger022: `ponytail` at the start looks like a stray word.

scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,171 @@
...
-    console.log(`Total current holders: ${holders.length}`);
-    if (cachePath !== "-") {
-      fs.writeFileSync(cachePath, JSON.stringify(holders, null, 2));
-      console.log(`Wrote holder cache: ${cachePath}`);
-    }
+  const file = HOLDERS_FILE ?? `.prime-holders-${chainId}.json`;
+  console.log(`${PRIME_DEPLOYMENT} ${dep.address} chainId ${chainId}, holders file ${file}`);
+
+  if (STAGE === "index") {
+    const fromBlock = FROM_BLOCK ?? (chainId === 56 ? PRIMEV2_DEPLOY_BLOCK_BSCMAINNET : 0);
```
Debugger022: Off mainnet this falls back to block 0, so a testnet index scans about 25k ranges instead of 3k. `deployments/bsctestnet/PrimeV2.json` is at 112385727. Worth a second constant, or requiring `FROM_BLOCK` off mainnet?

scripts/prime-update-scores.ts
```diff
@@ -1,272 +1,171 @@
...
-    holders = eventsToHolders(events);
-    console.log(`Total current holders: ${holders.length}`);
-    if (cachePath !== "-") {
-      fs.writeFileSync(cachePath, JSON.stringify(holders, null, 2));
-      console.log(`Wrote holder cache: ${cachePath}`);
-    }
+  const file = HOLDERS_FILE ?? `.prime-holders-${chainId}.json`;
+  console.log(`${PRIME_DEPLOYMENT} ${dep.address} chainId ${chainId}, holders file ${file}`);
+
+  if (STAGE === "index") {
```
GitGuru7, line 135: Should we only allow the index when `pendingScoreUpdates > 0`? Holders can only join or leave when no round is running, since the contract rejects these actions during an active round.

GitGuru7:
overall lgtm

scripts/prime-update-scores.ts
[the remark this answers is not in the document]
fred-venus: Enforced in b2d2e1ff instead of documented. While pendingScoreUpdates > 0 all six mint/burn entry points revert ScoreUpdateInProgress, so the holder set is frozen for the whole round and the only stale window is index → round open. STAGE=index now records nextScoreUpdateRoundId in the file and STAGE=update throws if the current round differs. Same round id ⇒ same holder set; equal-count churn can't slip through.

scripts/prime-update-scores.ts
fred-venus: Done — count check first, file written only if it passes.

scripts/prime-update-scores.ts
fred-venus: You're right, the header was wrong about mid-round changes. Rewritten: the list can't go stale during a round; index pins the round id, update refuses any other; an open round blocks the monthly burnBatch/issueBatch. Re-scan dropped.

scripts/prime-update-scores.ts
fred-venus: No reason — restored Promise.all.

scripts/prime-update-scores.ts
fred-venus: Removed.

scripts/prime-update-scores.ts
fred-venus: Dropped the constant; FROM_BLOCK now defaults to dep.receipt.blockNumber from the hardhat-deploy artifact (bscmainnet 107,040,035 / bsctestnet 112,385,727), so every network starts at the proxy deploy block.

scripts/prime-update-scores.ts
fred-venus: Same guarantee, slightly looser: index records nextScoreUpdateRoundId and update refuses a file from any other round. Every round opening bumps the id and holders are frozen while one is open, so matching ids ⇒ identical holder set. Indexing before the round opens is still allowed (useful to rehearse), the file is just rejected until re-indexed. Happy to hard-require pendingScoreUpdates > 0 at index time too if you prefer.

GitGuru7 (approved):

Debugger022 (approved):

PR #710, opened 2026-09-04
[VPD-1982]: support PrimeV2 markets with underlyings above 18 decimals

Debugger022:
## Description

Resolves VPD-1982

Prime normalised a market's capital to 18 decimals by multiplying by `10 ** (18 - decimals)`, which underflows for a wider underlying. `addMarket` therefore rejected those underlyings outright, via the guard added for audit finding **I06**. Capital now scales down instead, and the guard's ceiling moves from 18 to 24.

This unblocks listing the Venus Hub receipt tokens `vvhUSDT`, `vvhUSDC` and `vvhU`. Their underlyings have 24 decimals.

Also adds `MarketFacet.enterMarketForAccount(address account, address vToken)`, an ACM-gated entry point that enters a market for an account without a delegate grant from that account. It exists for `HubRouter` (VenusProtocol/venus-liquidity-hub#24), which supplies a vh position on the user's behalf and needs it to count as collateral in the same call; `enterMarketBehalf` requires the account to have approved the caller as a delegate first, which defeats a one-call supply. The role string is the function signature, `enterMarketForAccount(address,address)`, and the caller can only ever enter a market for an account, never exit one or move funds. The Prime and Comptroller changes are independent; they ship together because both gate the same vh market listing.

Note for reviewers, since the ticket described it differently: the underflow was **not** reachable on `develop`. The I06 guard rejected such a market before any scoring ran, so `addMarket(vvhUSDT, ...)` reverted with `UnsupportedUnderlyingDecimals(24)`, not a panic. Both halves of this PR are required; neither alone changes anything observable.

### Why 24 and not higher

The oracle reports price as `usd × 10 ** (36 - decimals)`. The further past 24 you go, the more that truncates: at 36 decimals it is integer dollars, so any sub-$1 asset prices to 0, and `ResilientOracle` treats 0 as `INVALID_PRICE` and reverts. That would reintroduce the same broken-market failure this PR removes. 24 is what the product needs and the width the tests actually prove scoring correct at.

### Changes

| File | Change |
| --- | --- |
| `PrimeV2.sol:1279` | Scale capital down above 18 decimals instead of underflowing |
| `PrimeV2.sol:723` | Ceiling `18 → 24`, plus natspec on the error and `addMarket` |
| `PrimeLens.sol:150` | Same scaling fix, so the frontend APR matches the on-chain score |
| `PrimeV2.decimals.ts` | New, 17 unit tests |
| `PrimeV2DecimalsForkTest.ts` | New, 18 bscmainnet fork tests |
| `PrimeV2.markets.ts` | Repin the existing I06 guard test from 19 to 25 decimals |
| `MarketFacet.sol:213` | New `enterMarketForAccount`, ACM-gated on its own signature, zero-address checked, reuses `addToMarketInternal` |
| `IMarketFacet.sol` | Declare it |
| `comptrollerTest.ts` | 6 new tests: enters without delegate approval, ACM revert, zero account, unlisted market, idempotent, `enterMarketBehalf` unchanged |

### Test plan

```bash
yarn test                                                                     # 1362 passing
FORKED_NETWORK=bscmainnet npx hardhat test tests/hardhat/Fork/PrimeV2DecimalsForkTest.ts   # 18 passing
```

Unit tests cover score arithmetic (equal value scores equally at 6, 8, 18 and 24; capped and uncapped; sub-unit dust truncates to zero), the `addMarket` boundary at 24 and 25, the Comptroller hook, `updateScores`, the XVSVault callback chain, and PrimeLens APR reads.

The fork tests upgrade the **live** PrimeV2 proxy at block 120600000 against real state (3 markets, 500 Prime holders, real oracle and vault) and confirm:

- the live implementation really does reject `addMarket(vvhUSDT)` with `UnsupportedUnderlyingDecimals`;
- after the upgrade the market registers, scores, and 1000 vhUSDT scores within 2% of 1000 USDT at real oracle prices;
- the Comptroller mint hook, XVSVault stake/unstake, and PrimeLens all work with the market registered;
- **no regression:** scores on the existing markets are recomputed across the upgrade and match **exactly** (vUSDT and vWBNB with real non-zero positions), aggregates are untouched, and `totalTokens`, alpha and the market list are preserved.

To confirm the unit tests are load-bearing, revert only the two scaling expressions while leaving the ceiling at 24: 14 of the 17 fail on the underflow, and the 3 guard tests correctly stay green.

Also verified: storage layout unchanged (37 slots, identical labels, offsets and type sizes, so the upgrade is safe); PrimeV2 at 23854 bytes with 722 of headroom; `prettier:check` clean and `solhint` reporting 0 errors.

### Deployment notes

**Ordering requirement — `initializeTokens` must run before `addMarket`.** The follow-on VIP in the ticket has these the other way round, which would break Prime.

`PrimeV2.accrueInterest(vvhUSDT)` calls `PLP.accrueTokens(vhUSDT)`, and vhUSDT is not initialised in the PrimeLiquidityProvider today (`lastAccruedBlockOrSecond(vhUSDT) == 0`). If the market is listed first, then for every Prime holder `updateScores`, the Comptroller mint hook, and `XVSVault.deposit`/`requestWithdrawal` all revert with `TokenNotInitialized(vhUSDT)`. `removeMarket` cannot back it out either, because it also calls `accrueInterest`. Only `PLP.initializeTokens([vhUSDT])` recovers it. Four fork tests pin this, including the recovery.

Correct order per market: `initializeTokens` → `addMarket` → `setTokensDistributionSpeed` → fund the PLP.

Other notes:

- **This reverses I06's remediation** from "reject > 18 decimals" to "handle up to 24". The auditor should see it.
- **Oracle wiring is already in place** — no VIP action needed. The live resilient oracle prices `vvhUSDT` at `1.0016e12`, exactly the `usd × 10 ** (36 - decimals)` convention, through the derived `CorrelatedTokenOracle` path. This matters because `ChainlinkOracle` and `BinanceOracle` both compute `18 - decimals` and cannot price anything wider than 18; any future wide-decimal listing must be checked the same way, since `addMarket` cannot verify it.
- **`addMarket` queues a score round for all 500 holders**, and the live `maxLoopsLimit` is 20, so the VIP needs **25 `updateScores` transactions** before `issue`/`burn` work again.
- **PrimeLens is a separate deployment** (`0x2f8c5e4562E22DB7908C56Bf99961C053436473c`) and needs its own upgrade.
- **Ship before the Prime `addMarket` VIP**, not alongside it.
- **`enterMarketForAccount` needs a diamond cut and a role grant.** The VIP must add the selector to the Core Comptroller's `MarketFacet` and then `ACM.giveCallPermission(comptroller, "enterMarketForAccount(address,address)", hubRouter)`. The liquidity-hub fork suite does both against the live Comptroller, with facet bytecode built from this PR, so that fixture has to be regenerated if this facet changes.
- **`Prime.sol` (V1) is deliberately out of scope.** Same expression at lines 714 and 846, and V1 is the only Prime on arbitrumone, basemainnet, ethereum, opmainnet, unichainmainnet and zksyncmainnet. No vh market is planned there. Recorded as a choice, not an oversight.

### DIFF
--- a/contracts/Comptroller/Diamond/facets/MarketFacet.sol
+++ b/contracts/Comptroller/Diamond/facets/MarketFacet.sol
@@ -202,6 +202,22 @@ contract MarketFacet is IMarketFacet, FacetBase {
     }
 
     /**
+     * @notice Add an asset to be included in an account's liquidity calculation, on that account's behalf
+     * @dev Entering a market makes the account's balance in it count as collateral, which also makes it
+     *  seizable in a liquidation
+     * @param account The address of the account entering the market
+     * @param vToken The address of the vToken market to enable for the account
+     * @return uint256 0=success, otherwise a failure (See enum Error for details)
+     * @custom:access Controlled by ACM
+     */
+    function enterMarketForAccount(address account, address vToken) external returns (uint256) {
+        ensureAllowed("enterMarketForAccount(address,address)");
+        ensureNonzeroAddress(account);
+
+        return uint256(addToMarketInternal(VToken(vToken), account));
+    }
+
+    /**
      * @notice Unlists the given vToken market from the Core Pool (`poolId = 0`) by setting `isListed` to false
      * @dev Checks if market actions are paused and borrowCap/supplyCap/CF are set to 0
      * @param market The address of the market (vToken) to unlist
--- a/contracts/Comptroller/Diamond/interfaces/IMarketFacet.sol
+++ b/contracts/Comptroller/Diamond/interfaces/IMarketFacet.sol
@@ -30,6 +30,8 @@ interface IMarketFacet {
 
     function enterMarketBehalf(address onBehalf, address vToken) external returns (uint256);
 
+    function enterMarketForAccount(address account, address vToken) external returns (uint256);
+
     function enterMarkets(address[] calldata vTokens) external returns (uint256[] memory);
 
     function exitMarket(address vToken) external returns (uint256);
--- a/contracts/Tokens/Prime/PrimeLens.sol
+++ b/contracts/Tokens/Prime/PrimeLens.sol
@@ -146,7 +146,10 @@ contract PrimeLens {
         aprInfo.borrowCapUSD = capital.borrowCapUSD;
 
         uint256 decimals = IERC20MetadataUpgradeable(_getUnderlying(market)).decimals();
-        uint256 scaledCapital = aprInfo.capital * (10 ** (18 - decimals));
+        // Normalize capital to 18 decimals. Underlyings with more than 18 decimals scale down.
+        uint256 scaledCapital = decimals <= 18
+            ? aprInfo.capital * (10 ** (18 - decimals))
+            : aprInfo.capital / (10 ** (decimals - 18));
 
         aprInfo.userScore = Scores._calculateScore(
             xvsStaked,
--- a/contracts/Tokens/Prime/PrimeV2.sol
+++ b/contracts/Tokens/Prime/PrimeV2.sol
@@ -167,7 +167,7 @@ contract PrimeV2 is
     /// @notice Error thrown when caller is not the PrimeLeaderboard contract
     error OnlyPrimeLeaderboard();
 
-    /// @notice Error thrown when underlying token decimals exceed 18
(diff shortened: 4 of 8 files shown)


fred-venus (approved):
lgtm

PR #711, opened 2026-09-17
chore: add code complexity checks to CI

GitGuru7:
Two CI gates, both enforced through existing scripts. No contract code changes.

## Code complexity

ESLint now caps per-function complexity at 15 for TypeScript:

- `complexity: ["error", 15]` — cyclomatic
- `sonarjs/cognitive-complexity: ["error", 15]` — cognitive, via `eslint-plugin-sonarjs@^1.0.4`

Solhint's `code-complexity` was already at `["error", 14]` and is untouched. 15 is a
ceiling, not a target, so a threshold that is already stricter stays where it is.

These run inside `yarn lint`, which CI already blocks on, so no workflow change was needed.

### What the gate found

Four functions were over the limit. One is fixed, three carry an inline
`eslint-disable-next-line` with a reason:

| Function | File | Cyclomatic | Cognitive | Action |
| --- | --- | --- | --- | --- |
| `generateCutParams` | `script/deploy/comptroller/facet-cut-params-generator.ts` | 33 → **4** | 56 → **3** | **Fixed** |
| `atomicLiquidate` | `scripts/bstock/atomic-liquidate.ts` | 49 | 58 | Suppressed |
| `buildSafeFallbackBatch` | `scripts/bstock/safe-fallback.ts` | 22 | 24 | Suppressed |
| `sweepOne` | `tests/hardhat/Fork/BStockLiquidatorFork.ts` | 29 | 30 | Suppressed |

`generateCutParams` was split into named pipeline stages — `resolveNewFacetAddresses`,
`resolveNewSelectors`, `identifyOldFacets`, `buildCutEntries`, `collectRemovals`. Pure
refactor: every statement is unchanged and runs in the same order.

The two bStock scripts are liquidation tooling and too important to reshape alongside a
lint rollout. They are suppressed here and get their own PR.

The fork test is low priority and does not need fixing.

## Storage layout and upgrade safety

New `Upgrade safety` workflow, pull requests only, running two checks:

moved to : https://github.com/VenusProtocol/venus-protocol/pull/712

## Slither: tried and reverted

Added in a9649027, reverted in e242e08f, kept in history so the next person does not repeat it.

It works, but the signal is not there. 382 findings out of the box. Filtering the vendored
0x fixed-point math library, the legacy contract directories and the known false-positive
detectors got it to 192 — still ~190 unreviewed items, and the high-impact ones were
dominated by patterns that are the architecture rather than bugs: every
`controlled-delegatecall` hit is a proxy doing delegatecall.

A Security tab nobody opens is worse than no Security tab. With AI-assisted development and
review in the loop, generic pattern matching of this kind adds little.

## Follow-ups

- Reduce `atomicLiquidate` and `buildSafeFallbackBatch` in their own PR.
- Solidity standards and review quality get better skills and smart-contract-specific
  review access to Trump, rather than another generic static analyser:
  - **Pashov Audit Group skills** — `solidity-auditor` (AI security audit), `x-ray`
    (pre-audit scan: threat model, invariants, entry points) and `fizz` (Echidna/Medusa
    fuzz suite generation). https://github.com/pashov/skills
  - **Historical findings lookup** — searchable 50k+ real audit findings.
    https://solodit.cyfrin.io
  - **Audited-primitives pin** — use installed audited code, never reinvent.
    https://www.openzeppelin.com/news/introducing-openzeppelin-skills
and more.


scripts/checkStorageLayout.ts
```diff
@@ -0,0 +1,380 @@
...
+ */
+function collectTargets(): Target[] {
+  const targets: Target[] = [];
+  for (const network of NETWORKS) {
+    const dir = path.join(ROOT, "deployments", network);
+    if (!fs.existsSync(dir)) continue;
+    for (const file of fs.readdirSync(dir).filter(f => f.endsWith(SUFFIX))) {
+      const key = `${network}/${file.slice(0, -SUFFIX.length)}`;
+      targets.push(targetFor(key, path.posix.join("deployments", network, file)));
+    }
```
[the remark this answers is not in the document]
GitGuru7: A deleted artifact is visible in the PR diff, and a reviewer should be asking why. The genuinely invisible case is a rename, which reads as delete+add and lets both halves pass free, this scripts scope does not covers the deleted case intentionally

scripts/checkStorageLayout.ts
```diff
@@ -0,0 +1,380 @@
...
+  | { kind: "failed"; detail: string };
+
+/** An allowlist entry only suppresses a failure; it never creates one. A target that is simply new
+ *  is passed before the allowlist is consulted, which is why new deployments must not be listed. */
+function verdictFor(target: Target, current: Map<string, Layout>, allowlist: Record<string, string>): Verdict {
+  if (target.isNew) return { kind: "new" };
+  const detail = target.blocked ?? check(target, current);
+  const allowed = target.key in allowlist;
+  if (detail) return allowed ? { kind: "skipped" } : { kind: "failed", detail };
+  return allowed ? { kind: "stale" } : { kind: "ok" };
```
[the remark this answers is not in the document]
GitGuru7: fixed: [dd56ca7](https://github.com/VenusProtocol/venus-protocol/pull/711/commits/dd56ca745796f0d11bf3f17b67cad2c723ccb61e)

script/deploy/comptroller/facet-cut-params-generator.ts
fred-venus, line 1: the change on this file is to reduce the complexity ?

package.json
```diff
@@ -29,6 +29,7 @@
     "lint:sol:fix": "prettier --write \"contracts/**/*.sol\"",
     "prettier": "prettier --write \"**/*.{js,json,md,ts,yaml,yml,sol}\"",
     "prettier:check": "prettier --check \"**/*.{js,json,md,ts,yaml,yml,sol}\"",
+    "prettier:cached": "prettier --write --cache \"**/*.{js,json,md,ts,yaml,yml,sol}\"",
```
Debugger022, line 32: Nothing calls this, and CI runs `prettier:check`. Leftover?

GitGuru7: This is intended to speed up Prettier locally. Usually, only the affected code needs to be formatted, while CI should continue running the full check as before.

The goal is to help us (dev) format their changes faster instead of running Prettier across the entire project. This should be especially useful in the VIP repo, where formatting the complete project takes a considerable amount of time unnecessarily.

script/deploy/comptroller/facet-cut-params-generator.ts
GitGuru7: Yes, and it remains the same overall. It’s just been restructured into smaller methods

fred-venus (approved):

PR #712, opened 2026-09-17
feat: check storage layout against deployed implementations in CI

GitGuru7:
## Storage layout and upgrade safety

New `Upgrade safety` workflow, pull requests only, running two checks:

- `yarn check:upgrade-safety` — OpenZeppelin `upgrades-core validate` over `build-info`:
  constructors, delegatecall, selfdestruct, immutables, initializers.
- `yarn check:storage-layout` — compares every proxied mainnet implementation against the
  layout recorded in its deployment artifact.

The reference layout is read from the PR **base branch**, so a deploy PR cannot pass by
overwriting its own reference. Covers seven mainnets: bscmainnet, ethereum, arbitrumone,
opbnbmainnet, basemainnet, opmainnet, unichainmainnet. Testnets are excluded because their
layouts drift on purpose, and zksync builds into a different build-info directory.
vTokens and Diamond facets are handled explicitly.

`scripts/storage-layout-allowlist.json` holds four bscmainnet entries, all the same case:
the deployment artifact is a stub with only abi, address and numDeployments, so there is no
reference layout to compare against. Every entry needs a reason, and an entry that starts
passing is reported as stale and fails the check, so the file cannot outlive the problem it
documents.


scripts/checkStorageLayout.ts
```diff
@@ -0,0 +1,390 @@
...
+  return Boolean(output.contracts?.[fqName.slice(0, at)]?.[fqName.slice(at + 1)]);
+}
+
+/** Compares one deployed layout against the current source, returning undefined when they are
+ *  compatible. A contract missing from build-info is a failure rather than a skip, so a source
+ *  that was renamed or deleted cannot drop out of the check unnoticed. */
+function check(target: Target, current: Map<string, Layout>): string | undefined {
+  const updated = target.fqName ? current.get(target.fqName) : undefined;
+  if (!updated) return `${target.fqName} was not found in artifacts/build-info`;
+  const report = getStorageUpgradeReport(target.deployed as Layout, updated, withValidationDefaults({}));
```
fred-venus, line 317: from ai:

this only compares same contract the version before and after right ? but for diamond pattern all the facets under the hood share the same contract storage layout.  so it could be facetA and facetB conflict as well, maybe we should also enforce the convention in CI to reject storage variables declared in facet contract, if really need anything put it into ComptrollerStorage.

The format could be a check specifically designed for Comptroller

GitGuru7: [90b5c7d](https://github.com/VenusProtocol/venus-protocol/pull/712/commits/90b5c7daba3b363bbf58d52059cea770078bfe71)

fred-venus (approved):

Debugger022 (approved):

Issue #713, opened 2026-09-18
ComptrollerLens._calculateSeizeTokens truncates to 0 when the borrowed asset has 24 decimals (vhUSDT/vhUSDC/vhU) — liquidator repays but seizes nothing


fred-venus:
## Summary

`ComptrollerLens._calculateSeizeTokens` divides before it multiplies:

```solidity
// contracts/Lens/ComptrollerLens.sol (develop, _calculateSeizeTokens)
Exp memory numerator   = mul_(Exp({ mantissa: liquidationIncentiveMantissa }), Exp({ mantissa: priceBorrowedMantissa }));
Exp memory denominator = mul_(Exp({ mantissa: priceCollateralMantissa }),     Exp({ mantissa: exchangeRateMantissa }));
seizeTokens = mul_ScalarTruncate(div_(numerator, denominator), actualRepayAmount);
```

`div_(Exp, Exp)` is `numerator * 1e18 / denominator`, so the intermediate ratio only keeps 18 decimals of precision and is truncated **before** `actualRepayAmount` is applied. For 18-decimal borrowed assets the ratio is large and the truncation is negligible. For the 24-decimal vhToken markets listed in VIP-659 (vhUSDT / vhUSDC / vhU, underlying decimals 24, price mantissa ≈ `1e12` = `10^(36-24)`) the numerator is `1e6` times smaller, and the ratio itself rounds down to a single-digit integer or to `0` whenever the collateral is a high-priced asset.

The result is that a liquidation of a vhToken **debt** seizes far less collateral than it should, or none at all, while the repayment still goes through.

## Current exposure

None today. All three vh markets have `borrowCap = 0` and the `BORROW` action paused (verified on BSC mainnet at block 122578443), and VIP-659 lists them as collateral-only. The path only becomes reachable if borrowing a 24-decimal (or any `priceMantissa << 1e18`) asset is ever enabled. Using a vhToken as **collateral** is unaffected: the same ratio for e.g. borrow vUSDT / seize vvhUSDT is ≈ `1.1e8` and rounds fine.

## Reproduction (read-only, BSC mainnet, block 122578443)

Production Comptroller `0xfD36E2c2a6789Db23113685031d7F16329158384` → `comptrollerLens()` = `0xd5DEb631cB6c6a667e926a482aadc95a471b120c` (verified source matches `develop`).

```bash
COMP=0xfD36E2c2a6789Db23113685031d7F16329158384
[...]

PR #604, opened 2025-07-09
[VEN-3321]: Liquidation Threshold and repay Improvements in Core Pool on BNB

Debugger022:
## Description

This PR implements the following features:-

[VEN-3321] - Add the concept of Liquidation Threshold to the core Pool on BNB.
[VEN-3322] - Dynamic Liquidation Incentive and Close Factor in core Pool on BNB.
[VEN-3323] - Define maximum Liquidation Incentive per seized asset in the core pool on BNB.
[VEN-233] - Repay logic improvisation

This pull request introduces significant enhancements and refactoring to the contracts, primarily to support dynamic liquidation incentives, improved health factor calculations, and integration with a new `LiquidationManager` contract. The changes include new interfaces and storage structures, updated function signatures, and expanded account health and liquidation logic.

**Liquidation and Account Health Enhancements**

* Added new functions and updated existing ones in `ComptrollerInterface.sol` to support dynamic liquidation incentives and close factor, health factor calculations, and detailed account snapshots, including overloaded methods for seize token calculations. 

**LiquidationManager Integration**

* Added references to the new `LiquidationManager` contract in both storage and interface files, and exposed its address via the Comptroller contract.

**Storage and Contract Refactoring**

* Deprecated old close factor and liquidation incentive variables, replacing them with new ones (`__oldCloseFactorMantissaSlot`, `maxLiquidationIncentiveMantissa`) and moved to `ComptrollerV19Storage`. Updated all relevant contract inheritance to use the new storage version. Also, removed the old `setCloseFactor` and `setLiquidationIncentive` functions as they will now be calculated dynamically using the `liquidationManager`.

**Diamond Facet and Functionality Updates**

* Updated `FacetBase` and `MarketFacet` to use new health snapshot logic and ComptrollerLens integration, including new internal and external functions for health factor calculations and liquidation seize token calculations.

**Repay logic Improvisation**

* Updated the `repayAmount` calculation in the `VToken` from [this](https://github.com/VenusProtocol/venus-protocol/blob/bdb945591e6b3ee4d483e77e2d7a74a2f754bfe0/contracts/Tokens/VTokens/VToken.sol#L1190) to this:
`vars.repayAmount = repayAmount >= vars.accountBorrows ? vars.accountBorrows : repayAmount;`

PR #714, opened 2026-09-22
fix(lens): compute seize tokens with a single division

Debugger022:
## Problem
`ComptrollerLens._calculateSeizeTokens` rounded the ratio `LI · priceBorrowed / (priceCollateral · exchangeRate)` to 18 decimals before multiplying by the repay amount. For 24-decimal debt (price mantissa ~1e12) this drops part or all of the seize amount. On BSC mainnet, repaying 100 vhUSDT against vBTC collateral seizes 0 while the repayment still goes through.

## Fix
- Single division: `repay · LI · priceBorrowed / (priceCollateral · exchangeRate)`, floored once.
- One lens swap fixes all three entry points (VToken, legacy vBNB, VAI).
- NatSpec corrected for the price scale and the return value.

## Behavior change
- All markets now seize the exact floor. On live 18-decimal pairs this is 0 to 0.11% more collateral at realistic repay sizes (up to 0.31% on a $1 repay), never less.
- Liquidations sized off-chain with the old formula right at the borrower's collateral balance may need a slightly smaller repay to avoid `LIQUIDATE_SEIZE_TOO_MUCH`.

## Tests
- Unit: mainnet vectors including vh↔vh pairs, decimals matrix {6, 8, 18, 24}, rounding boundaries, seeded random cases, reverts, all entry points.
- Integration: real VToken and Liquidator contracts (24→18, 24↔24, 18↔18, max-seize boundary, dust).
- Fork (block 123190984): every live pair against the deployed lens, the VAI entry point, account-liquidity parity, and end-to-end liquidations through the production Liquidator with the old and new lens.
- Run the fork suite with `FORKED_NETWORK=bscmainnet npx hardhat test tests/hardhat/Fork/liquidationSeizePrecision.ts`.

## Not in this PR
- Zero-seize guard in `VToken.liquidateBorrowFresh`. After this fix a zero seize only happens for dust, which already seizes 0 on 18-decimal markets today.
- Isolated pools (same pattern), and deployment/VIP of the new lens.
- Conflicts with #604 (same function). Keep this body when resolving.

Refs #713


fred-venus (approved):
lgtm

GitGuru7 (approved):
lgtm

